PR #2, opened 2015-07-23
Enable connecting build-image(s) to releases

ycheng-aa:
1. Connect build-image to releases. They have many to many
relationship now.
2. Change old build-image's release and version to rpm_release
and rpm_version. Now use latter to filter build images.


erichuanggit:
looks good to me.

lubomir:
Merged as 8f08b1b9.

PR #4, opened 2015-07-27
Refactor resource url field to id in contact app

erichuanggit:
Jira: PDC-883

### DIFF
--- a/pdc/apps/contact/serializers.py
+++ b/pdc/apps/contact/serializers.py
@@ -19,10 +19,7 @@ class ContactRoleSerializer(StrictSerializerMixin,
 
     class Meta:
         model = ContactRole
-        fields = ('url', 'name')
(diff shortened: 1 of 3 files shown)


ycheng-aa:
Looks good.

lubomir:
Exposing the `id` for contact roles does not seem very useful to me: roles are identified by the name, so knowing the `id` will give the user no new option. In my opinion contact roles should either keep the URL or only consist of the name.

Other than this, it looks good to me.

erichuanggit:
All right, I will keep only consist of name for contact roles.

lubomir:
With the change it looks good to me. Could you try rebasing the branch on master and force pushing to GitHub? (I would like to avoid to history with merge commits all over the place.)

erichuanggit:
I will do that.

PR #6, opened 2015-07-27
Refactor resource url field to id in component app

erichuanggit:
Known issues: still keep the url in nested resource, for instance:
* contact url in release_component and global_component

JIRA: PDC-884


lubomir:
Looks good to me.

PR #7, opened 2015-07-27
Enable continuous integration on Travis-CI

lubomir:
After every pull requests, tests will be executed and the results
displayed in GitHub interface. We could drop this after we integrate
with our Jenkins instance, or keep using Travis instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: python
+python:
+    - "2.7"
+addons:
+    apt:
+        packages:
+            - uuid-dev
+            - swig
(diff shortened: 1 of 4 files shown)


lubomir:
Ok, this won't be as easy as I though. The tests are failing due to koji not being installed. Since Travis runs on Ubuntu, I have no idea how to install koji there.

lubomir:
I've managed to come up with a hack that will allow the tests to be run and actually even succeed. Right now the integration with Travis-CI is turned off, I will turn it on again after this is merged.

PR #10, opened 2015-07-28
Add id to override rpms

lubomir:
This pull request adds identifier to override rpms. This allows easier deleting.

Fixes PDC-683 and PDC-885.


erichuanggit:
looks good to me.

PR #14, opened 2015-07-29
Improve compose images

lubomir:
* document `compose` filter
* display a list of composes for each image

PDC-712


erichuanggit:
:+1:

PR #15, opened 2015-07-30
Bulk non-standard method call

lao605:
trap calls using methods which are not standards in HTTP/1.1 in the middleware.

### DIFF
--- a/CONTRIBUTING.markdown
+++ b/CONTRIBUTING.markdown
@@ -1,0 +1,73 @@
+# Contributing to PDC
+
+We welcome all contributions from bug fixes to documentation improvements. If
+you want to help, feel free to open a pull request.
+
+
+## Pull Requests
+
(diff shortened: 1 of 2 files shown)


lubomir:
Since this pull request is from a forked repository, checking this change out locally is not completely straightforward. The way I did it is to use this command `git fetch origin pull/15/head:bulk-methods`, which creates a local branch `bulk-methods` with all commits from pull request #15. After playing around with the change and testing it I can delete the local branch without leaving anything behind.

lao605:
wow, Travis CI seems to work pretty well~ @lubomir

lubomir:
I don't have any more comments, this change looks good to me. Since there is some weird merging going on in these commits, I squashed and rebased them to simplify review. It is available on `PDC-758-wrong-method` branch as commit 6a84272.

lubomir:
Merged as 6a84272.

PR #16, opened 2015-07-30
Mark key_id in SigKey resource as read_only

lubomir:
Django REST Framework did not allow updating fields that were present in
URL in version 2.4.*. In current version, it is allowed. Since we don't
want to allow users to change key_id, we have to do so explicitly.

There is a test that tries to patch the key id. The expected behaviour
is that no error will be reported, but no change will happen.

JIRA: PDC-634


erichuanggit:
:+1:

PR #17, opened 2015-07-31
Expose information about current user in API

lubomir:
This patch adds support for obtaining information about current user
through the API. This can not be used to spy on other users as there is
no input.

Tests are included.

JIRA: PDC-811


erichuanggit:
:+1:

PR #20, opened 2015-08-04
Add a new Messenger using `qpid.messaging`.

xychu:
Which supports mutli-endpoints connection failover;
and also has connection timeout and msg send timeout.


xychu:
Not needed anymore.

PR #21, opened 2015-08-04
Update documentation to reflect newer productmd

lubomir:
Since we now depend on (currently unreleased) productmd, we need to
update some parts of the code.

This patch updates all mentions of compose ordering in documentation as
well a comment in a test in release app. The test continues to work, the
comment is updated to better explain what is going on.

The remaining parts where productmd is used were all updated before the
open-sourcing. These are release/compose/rpms/image import and exporting
rpm manifest.

There are no code changes, just documentation and comments.


simozhan:
thumbup

PR #23, opened 2015-08-05
PDC-645 Refactor compose/package

ycheng-aa:
Add 2 new end points to take old compose/package
endpoint function.
One is  for a given release and srpm_name list all
composes that contain the package (and include its version).
Another is given a compose _C_ and a package, find the
latest compose older than _C_ which contains a
different version of the package.
Keep the old endpoint same for a while to don't break
existing scripts.

### DIFF
--- a/pdc/apps/compose/tests.py
+++ b/pdc/apps/compose/tests.py
@@ -172,6 +172,119 @@ class VersionFinderTestCase(APITestCase):
                          [{'compose': 'compose-3', 'packages': ['bash-0:5.6.7-8.x86_64.rpm']}])
 
 
+class FindComposeByReleaseRPMTestCase(APITestCase):
+    fixtures = [
+        "pdc/apps/common/fixtures/test/sigkey.json",
+        "pdc/apps/package/fixtures/test/rpm.json",
+        "pdc/apps/release/fixtures/tests/release.json",
+        "pdc/apps/compose/fixtures/tests/variant.json",
+        "pdc/apps/compose/fixtures/tests/variant_arch.json",
+        "pdc/apps/compose/fixtures/tests/compose_overriderpm.json",
+        "pdc/apps/compose/fixtures/tests/compose.json",
+        "pdc/apps/compose/fixtures/tests/compose_composerpm.json",
+        "pdc/apps/compose/fixtures/tests/more_composes.json",
+    ]
+
+    def test_get_for_release(self):
+        url = reverse('findcomposebyrr-list', kwargs={'rpm_name': 'bash', 'release_id': 'release-1.0'})
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data,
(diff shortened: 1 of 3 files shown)


lubomir:
Couple notes:
* generally this looks good
* the APIs don't show up in API root, so they are hard to discover
* I think we should drop the `to_dict` option by picking one behaviour and sticking with it

ycheng-aa:
@lubomir The "I think we should drop the to_dict option by picking one behaviour and sticking with it", you means drop to_dict only for these 2 new endpoints or for all endpoints including compose/package?

lubomir:
@ycheng-aa Just here, we should keep compose/package unmodified for backwards compatibility. Just simplify the new APIs. I don't think there are any business reason why someone should need this option, we just need a good default.

ycheng-aa:
@lubomir As discussed, I didn't add these 2 end points to API root page issue. And set to_dict to False first for these 2 in 99828363.

lubomir:
The documentation on `FindLatestComposeByComposeRPM` is a bit imprecise: the view always returns a single compose, so there is no need to describe what the ordering is, and the return value is always a single JSON object. The `latest` argument also does not make much sense there, and `to_json` is forgotten ([line 1288](https://github.com/release-engineering/product-definition-center/commit/998283636bf5cdaee801a2cc135d60952b727652#diff-5526c0e525542969f5493b60102e5a4aR1288)).

ycheng-aa:
@lubomir  Fixed in ada1f64bc5f2. Please review.
My original understanding is the which one is 'latest' should depend on the order that described. First define the order, then pick the latest one according to the order. If not like this, remove the order part is good. Anyway, I removed it in this commit.

lubomir:
@ycheng-aa You are right, the order still determines which compose is latest. Maybe it needs to be documented, I'm not sure.

ycheng-aa:
@lubomir Ok, I reverted the documentation about the "order" for FindLatestComposeByComposeRPM.

sochotnicky:
@lubomir FWIW the to_dict was requested earlier to simplify figuring out exact NVR bits without having to parse it

ycheng-aa:
@sochotnicky @lubomir So the final decision is keep "to_dict" or what?

erichuanggit:
I would agree with stano. let's keep the original parameters since all these requirements  were coming from clients.

ycheng-aa:
@lubomir  @erichuanggit  I add to_dict back for code/tests/documentation and rebased it. Please help me review it. Thanks.

lubomir:
Looks pretty good to me. I would prefer the commit title to not start with the PDC-645 and instead for it to be somewhere in the commit message. It is not very useful for people without access to Jira, so it does not need such prominent place.

ycheng-aa:
I will change the commit message in final rebase. The coverage failed seems because messaging. It should not be related to this pull request.

erichuanggit:
Looks good to me.

lubomir:
:+1:

PR #25, opened 2015-08-06
Get correct version by git describe

lubomir:
When the pdc module is imported into another git repo, the version
reported is the version from that repo, not from PDC.

### DIFF
--- a/pdc/__init__.py
+++ b/pdc/__init__.py
@@ -6,8 +6,10 @@
 import os
 import subprocess
 
-VERSION = "0.5.0-0.1.beta"
+VERSION = "v0.1.0-alpha.2"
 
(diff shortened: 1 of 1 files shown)


pdc/__init__.py
```diff
@@ -8,9 +8,11 @@
 
 VERSION = "0.5.0-0.1.beta"
```
erichuanggit: :+1: 
how about update this version to current upstream release version "v0.1.0-alpha"?

lubomir:
@erichuanggit I will update the version. Since the first tag I create was not annotated, I also created a new test release with annotated tag.

erichuanggit:
:+1:

PR #27, opened 2015-08-06
Add queryset to current user API

lubomir:
If server is configured to use model permissions, each view set must
have a queryset so that there is a place from where to get the
permissions.

### DIFF
--- a/pdc/apps/auth/views.py
+++ b/pdc/apps/auth/views.py
@@ -7,7 +7,7 @@
 import json
 
 from django.shortcuts import render, redirect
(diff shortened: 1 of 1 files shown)


erichuanggit:
:+1: 
It would be better if we have unit test to cover this case.

lubomir:
We should eventually update all tests to check for authentication. There already are tests for the current-user API, and when server is configured to use model permissions, they fail. We just can't run the tests with that settings as all tests that connect to API are unauthenticated and would fail.

PR #28, opened 2015-08-06
Return labels connected to global components

lubomir:
When labels are added to a global component, the response now includes
details about the label instead of being empty.

Both test cases for this feature were updated to verify the response is
corrent. The label serializer also includes a URL in the response, but
the URL is not checked to be correct.

JIRA: PDC-566, PDC-768, PDC-777


erichuanggit:
looks good to me.

simozhan:
it looks good

PR #29, opened 2015-08-06
Log exception when 503 is returned

lubomir:
This should help with troubleshooting problems on devel or qa servers as
the traceback will be in server error log.

There is a change in bulk ops test suite. Some of the tests were
actually wrong and raised an exception for a wrong reason. All exception
logging in those tests is silenced.

This can help development since if there is an unhandled exception
causing 503, the tests now show a traceback of the error instead of just
the failed assertion. There is no change when looking at the bad page in
browser.

### DIFF
--- a/contrib/bulk_operations/tests.py
+++ b/contrib/bulk_operations/tests.py
@@ -3,6 +3,8 @@
 # Licensed under The MIT License (MIT)
 # http://opensource.org/licenses/MIT
 #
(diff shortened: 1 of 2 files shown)


erichuanggit:
:+1: 
vote for this patch, generally I will comment out "EXCEPTION_HANDLER" in settings file in local development :-)

PR #32, opened 2015-08-10
Fix internal server error on OPTIONS requests

lubomir:
The metadata generator assumed that when a view responds to a PUT
request, it must be possible to call `get_object` on it to test object
permissions. This is however not the case with bulk operations, which
add PUT to list url. An unhandled AssertionError was raised in that
case.

This patch provides a custom BulkMetadata generator, which is pretty
much the same as the one provided by REST Framework, only it is aware
of the bulk operations.

The settings need to be updated in order for the generator to be used.
All settings files in this repository are updated, but anyone using
settings_local.py need to modify it accordingly.

### DIFF
--- a/contrib/bulk_operations/metadata.py
+++ b/contrib/bulk_operations/metadata.py
@@ -1,0 +1,68 @@
+#
+# Copyright (c) 2015 Red Hat
+# Licensed under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+#
+"""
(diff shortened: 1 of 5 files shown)


erichuanggit:
Looks good to me. 
I think we need to add the settings change to next deploy guide.

lubomir:
The deploy guide has nothing on overriding REST_FRAMEWORK settings, so I think this change should be picked up without any manual work..

PR #33, opened 2015-08-10
Fix quotes in documentation of build-images

lubomir:
In JSON, strings must be wrapped in double quotes. The list of
serializer fields is not actually JSON, but for consistency it makes
sense to use double quotes everywhere.

JIRA: PDC-854


simozhan:
It makes sense for consistency. +1

ralphbean:
Simple, yes.  :+1:

PR #35, opened 2015-08-10
Don't delete group type with a group

lubomir:
When there is a group of certain type, make it impossible to delete the
type. The original behaviour was incorrect: the cascading deletes would
not be logged in changeset. Disabling the cascade is both easier and
more practical for the user as deleting the type should be rare.

JIRA: PDC-875


simozhan:
Yes. Deleting the type should be rare. It is good enough to protect grouptype in ReleaseComponentGroup. +1

erichuanggit:
Looks good to me.

PR #36, opened 2015-08-11
Add StompMessenger to send msg via STOMP.

xychu:
Adds sample config in `settings.py`;
Requires `stomp.py` to be installed.

### DIFF
--- a/pdc/apps/utils/__init__.py
+++ b/pdc/apps/utils/__init__.py
@@ -5,15 +5,16 @@
 #
 from django.conf import settings
 
-from .messaging import DummyMessenger, KombuMessenger, FedmsgMessenger, ProtonMessenger
+from .messaging import (DummyMessenger, KombuMessenger, FedmsgMessenger,
+                        ProtonMessenger, StompMessenger)
 
+
+MESSENGERS = {
+    'kombu': KombuMessenger,
+    'fedmsg': FedmsgMessenger,
+    'proton': ProtonMessenger,
+    'stomp': StompMessenger,
+}
 
 # init messenger
-if settings.MESSAGE_BUS['MLP'] == 'kombu':
-    messenger = KombuMessenger()
-elif settings.MESSAGE_BUS['MLP'] == 'fedmsg':
-    messenger = FedmsgMessenger()
-elif settings.MESSAGE_BUS['MLP'] == 'proton':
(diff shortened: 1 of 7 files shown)


pdc/apps/utils/__init__.py
```diff
@@ -5,7 +5,8 @@
 #
 from django.conf import settings
 
-from .messaging import DummyMessenger, KombuMessenger, FedmsgMessenger, ProtonMessenger
+from .messaging import (DummyMessenger, KombuMessenger, FedmsgMessenger,
+                        ProtonMessenger, StompMessenger)
```
erichuanggit: I think we won't support `proton`. It's better to remove it, also in spec, requirement and settings. Otherwise it will mislead users.

pdc/apps/utils/messaging.py
```diff
@@ -49,3 +53,36 @@ def send_message(self, topic, msg):
...
+            self.connection.start()
+            self.connection.connect()
+        except Exception, e:
+            logger.warn("StompMessenger connection exception(%s): %s." % (type(e), e))
+            return False
+        return True
+
+    def send_message(self, topic, msg):
+        address = '/topic/' + settings.MESSAGE_BUS['TOPIC'] + str(topic)
+        if self.connected or self.do_connect():
```
erichuanggit: one of them should be enough.

lubomir:
As for the decreased test coverage, I think we should just ignore it. Unless we provide a mock messaging server to the test, we can not reasonably test the messengers themselves.

pdc/apps/utils/__init__.py
xychu: ProtonMessenger just like KombuMessenger, it works, just not support multi endpoints failover, so I suggest we leave it there but remove all related items from spec and requirements.

pdc/apps/utils/messaging.py
xychu: This is ugly, what I want to do is to try to `do_connect` again if it's not connected before sending, so that we could know that the server has been back alive.

xychu:
The `_request` will block message middleware, so I included it in this PR. 
And I'll change the `__init__` to make code more simple.

erichuanggit:
Looks good to me.

PR #37, opened 2015-08-11
Fix bulk operations tests.

xychu:
When use mock(v1.3.0), there is no `assert_called` for Mock object,
update three tests to make them pass.


simozhan:
OK for me. +1

erichuanggit:
met the same issue with mock 1.3.0, but I did not find `assert_called` in 1.0.1 either.
:+1:

lubomir:
This looks good. The original tests worked fine for me on mock 1.0.1, but this looks more bullet-proof.

PR #38, opened 2015-08-11
Fix issue that create component-groups without release components

erichuanggit:
Without release components provided, the default components is None
which can't be iterated. Update it to empty list.

JIRA: PDC-872


simozhan:
Cool. +1

lubomir:
Looks good to me.

PR #40, opened 2015-08-11
Handle bad data for bulk delete

lubomir:
When user tries to bulk delete data with dicts instead of identifiers,
we should report a sensible error instead of crashing the server.

JIRA: PDC-910


erichuanggit:
Looks good to me.

bliuredhat:
Good for me

PR #41, opened 2015-08-11
Document id of contact in role-contacts API

lubomir:
Add description of the nested identifier in role contacts API. The id
may be useful to some users, even though it is not immediately obvious
whether the id is coming from persons or maillists. This can be deduced
from the keys in the contact.

There should be no possibility of mistake to delete a person instead of
a maillist (or vice-versa). These models are subclasses of the same
parent and should share the same pool of identifiers.

JIRA: PDC-904


erichuanggit:
Looks good to me.

PR #42, opened 2015-08-12
Fix setting DEFAULT_METADATA_CLASS

simozhan:
Change DEFAULT_METADATA_CLASS to contrib.bulk_operations.metadata.BulkMetadata


erichuanggit:
Looks good to me.
Interesting all the other settings files `settings_local.py.dist, settings_test.py` point to the correct metadata. We were lucky that it was override by local settings.

xychu:
LGTM.
It should be a typo.

lubomir:
Yes, that is a typo, thanks for catching this.

PR #43, opened 2015-08-12
Make easy understand for ReleaseComponent fields input.

erichuanggit:
It was not friendly only with id for ReleaseComponent, instead provide
two option: one is object id, the other is resource's unique_together
fields. Both of them need to be input as a dict.
Impact endpoints include: component-groups and release-component-relationships

JIRA: PDC-905

### DIFF
--- a/pdc/apps/component/serializers.py
+++ b/pdc/apps/component/serializers.py
@@ -440,10 +440,21 @@ class ReleaseComponentRelatedField(serializers.RelatedField):
         return result
 
     def to_internal_value(self, data):
+        if not isinstance(data, dict):
+            raise serializers.ValidationError({'detail': "Input [%s] for ReleaseComponent must be a dict." % data})
+        kwargs = dict()
+        if 'id' in data:
+            kwargs['id'] = data.get('id')
+        else:
+            kwargs['release__release_id'] = data.get('release')
+            kwargs['global_component__name'] = data.get('global_component')
+            kwargs['name'] = data.get('name')
         try:
-            rc = ReleaseComponent.objects.get(id=data)
+            rc = ReleaseComponent.objects.get(**kwargs)
         except ReleaseComponent.DoesNotExist:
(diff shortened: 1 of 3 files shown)


lubomir:
Generally this looks good to me. I would like it if there were more error handling. It doesn't warn when extra fields are included in the dict specifying the component. It will accept the following without any error: `{"id": 1, "release": "hello-1.0", "name": "bash", "global_component": "bash"}` (in this case anything but `id` is ignored, but that is not obvious).

erichuanggit:
I admit that it was a lazy solution not causing application crash as some un-predict exception happens. There was a case in my mind, which also mentioned in your second comment, but a little different. with all fields input or at least `id` and its unique_together fields, but the value of `id` is not same as this resource id, `DoesNotExist` can't catch this exception. 
Anyway, I will put more error handling.

erichuanggit:
Haven't rebase the code. I will do it when review pass from team.

lubomir:
I'm not sure if it should be even more strict and report error on any unrecognized key. It will still accepts these wrong inputs:
```json
{"foo": "bar"}
{"iid": 1, "release": "foo"}
```
with message saying it can't find a release component.
This is still invalid, but will be accepted (release is ignored):
```json
{"id": 1, "release": "foo"}
```

erichuanggit:
that will need more checks, thinking about it.

lubomir:
:+1:

erichuanggit:
Looks more elegant.

simozhan:
Just two questions.
1. Is it ok to create relationship between id:1 and id:1?
2. Is it ok to create relationship between id and unique together

erichuanggit:
@simozhan what relationship?

lubomir:
@simozhan
1. There is probably no point in putting a component in a relationship with itself. But I don!t know if it needs a validation.
2. Yes, those are just two ways of specifying which component to use, they should be interchangeable.

PR #44, opened 2015-08-12
Fix bulk metadata generator in production

lubomir:
The AssertionError can only be raised in debug mode. On product server,
the assert is skipped and a KeyError is raised instead.


erichuanggit:
:+1:

PR #46, opened 2015-08-12
Order resources on web/cli

bliuredhat:
Order resources on web/cli.
To make an order for API is sorted by its first letter .The way is to rewrite the get_api_root_view() function in /contrib/bulk_operations/bulk_operations.py doc.

### DIFF
--- a/pdc/apps/utils/SortedRouter.py
+++ b/pdc/apps/utils/SortedRouter.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2015 Red Hat
+# Licensed under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+#
+
+from collections import OrderedDict
+from django.core.urlresolvers import NoReverseMatch
+from rest_framework import views
+from rest_framework.reverse import reverse
+from rest_framework.response import Response
+from contrib.bulk_operations import bulk_operations
+
+
+def get_resolver_match(request):
+    try:
+        return request.resolver_match
+    except AttributeError:
+        from django.core.urlresolvers import resolve
+        return resolve(request.path_info)
+
+
+class PDCrouter(bulk_operations.BulkRouter):
+    """ Order the api url  """
+
+    def get_api_root_view(self):
+        """
+        Return a view to use as the API root.
+        """
+        api_root_dict = OrderedDict()
+        list_name = self.routes[0].name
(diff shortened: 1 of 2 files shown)


lubomir:
The implementation looks fine to me. However, I think the files from your IDE should not be part of this commit. There is also an extra file called `out.html`.

Why the change in requirements? In my opinion it should be at least mentioned in the commit message or be separate commit altogether.

The merge conflict is caused by the change in settings. This commit should not change the settings as the change is unrelated to this task.

I don't think this should be part of `BulkRouter`, as it is not relevant to bulk operations. It might be better to create `PDCRouter` (or any better name) in `pdc/routers.py` (or maybe completely separate file) which would inherit from `BulkRouter` and implement this feature.

bliuredhat:
lubomir, Thank you so much. Your opinion is right and it's easy to edit.

lubomir:
You were right about changing the dependency. That will fix the build failure. Just mention it in commit message that the the router does not work with REST Framework 3.2.

bliuredhat:
lubomir , thank you for your help, it's good for me.

lubomir:
Looks great!

Since we already depend on Django >= 1.8.1, you can remove the workaround for older version.
If you want to, you could even write a simple test that will get API root and verify that the keys are really sorted (there is no need to check if all registered views are included). This could be in a separate commit and pull request and does not need to block this one.

bliuredhat:
lubomir, it sounds cool !

erichuanggit:
I would suggest moving the changes into a single file as what `BulkRouter` is, keeping `routers.py` tidy.

bliuredhat:
erichuanggit, It's cool ideal , thank you and I will do it.

lubomir:
Maybe `pdc/apps/utils/routers.py` would be better name. It would change the import into `from pdc.apps.utils.routers import PDCRouter`.

xychu:
+1

erichuanggit:
How about `SortedRouter.py`? `BulkRouter` looks like not making sense.

bliuredhat:
erichuanggit, lubomir ; Okay, it's a good suggest for me, and I will to do now.

pdc/apps/utils/BulkRouter.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
```
erichuanggit: `BulkRouter.py` this file is not required.

PR #48, opened 2015-08-12
Rename repos resource to delivery-targets.

ycheng-aa:
1. Add new endpoints with same function as repos related functions.
2. Change correspoinding documentation and add deprecated message.

Didn't change repos related fields in query and response.

### DIFF
--- a/pdc/apps/repository/views.py
+++ b/pdc/apps/repository/views.py
@@ -27,7 +27,10 @@ class RepoViewSet(ChangeSetCreateModelMixin,
                   mixins.RetrieveModelMixin,
                   viewsets.GenericViewSet):
     """
-    An API endpoint providing access to repositories.
+    An API endpoint providing access to content delivery repositories.
+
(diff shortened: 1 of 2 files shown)


lubomir:
All the URLs are missing trailing slashes. There will be a redirect, but the documentation should be precise.

I also feel the serializer field and filter should be renamed, but will write more about this in JIRA so that we can discuss it.

ycheng-aa:
Fixed the URL trailing slashes in ad6bdfb8c. Thanks.

dmach:
"delivery-target(s)" is not a good and descriptive name for the resources previously referring to "repos".
My preference is "content-delivery-repos" or "content-delivery-repositories".

ycheng-aa:
Has been changed to content-delivery-repos

erichuanggit:
Looks good to me.

lubomir:
Merged as c73310cf.

PR #49, opened 2015-08-12
Improve error message for wrong relationship type

lubomir:
Since the types are not exposed in API currently, we should at least
list them in error message when bad type is given.


erichuanggit:
Looks good to me.

erichuanggit:
PDC-892 could be linked to this patch.

PR #51, opened 2015-08-13
Able to set release_component_type to null.

erichuanggit:
JIRA: PDC-888


lubomir:
I don't think this is the right fix. Based on discussion with @sochotnicky on IRC, type should not even be optional and there should rather be a default value (most likely RPM).

erichuanggit:
Basically fixed by this bug's description. I will keep this change here until we meet together in next meeting.

lubomir:
:+1:

simozhan:
Good.

PR #52, opened 2015-08-13
Pdc 57

lao605:
make a pull request to enable better discussion. feel free to give any suggestion.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,11 +7,11 @@
 # docker build -t <YOUR_NAME>/pdc <the directory your Dockerfile is located>
 # 
 # 2. Running the container
-# 	2.1 To operate on the container interactively (with a terminal)
-# 	docker run -it -p 10000:8000 <YOUR_NAME>/pdc
+# 	2.1 To display the log interactively (with a terminal)
+# 	docker run -it -P -v $PWD:$PWD <YOUR_NAME>/pdc python $PWD/manage.py runserver 0.0.0.0:8000
 # 
 # 	2.2 To run the container in daemon mode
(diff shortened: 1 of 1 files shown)


erichuanggit:
could I try it on Fedora 20?

lao605:
the guide is misleading I will refine it today

lao605:
still welcome criticism

lubomir:
I don't really have criticism, just some questions and suggestions.

* Why is it downloading the code from github? Why not take the code from the same checkout where the Dockerfile is? This would allows testing changes that are not published yet, you would just rebuild the image.
* I tried building the image and running it, which worked fine. I'm not sure how else I could use it to help me. I can see the advantage of not having to set up dependencies in virtualenv and instead running test server in container, but if it means manually rebuilding and restarting the server each time a change is made it does not seem very desirable.
* Why does it install virtualenv at the end? It seems unused in the docker build.
* I would like to see the ability to connect to a PostgreSQL database running in another container. Maybe [docker-compose](https://docs.docker.com/compose/) might be handy here.

lao605:
@lubomir 
I was quite confused about the requirement and kind of considering developers will move their development job totally to the docker container(the "optional ipython notebook" kind of imply that and I wrongly assume everyone is using vim in this environment) so it just clones the project and hardly interact with the host. so maybe this is what is right: `1. a developer clone the project in his host`-->`2. developer build the container from the image which contains all the env setup and dependencies`-->`3. the developer changes code in his own computer and see the corresponding result in the container or directly in his pycharm console`. So dependencies and env setup is totally unnecessary in the host and the development process in friendly to both old and new developers because all the inconvenient dirty work is done by docker and developer still uses the editor/IDE they like.

PR #54, opened 2015-08-14
Add doc for messaging.

xychu:
Including:
* overview & dataflow
* supported messengers and sample settings
* to be improved

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -17,6 +17,7 @@ Contents:
    versioning
    deployment
    bulk_operations
+   messaging
 
 
 Indices and tables
--- a/docs/source/messaging.rst
+++ b/docs/source/messaging.rst
@@ -1,0 +1,195 @@
+.. _messaging:
+
+Messaging
+=========
+
+Overview
+--------
+
+Messaging enables PDC to send out useful informations to message bus so that other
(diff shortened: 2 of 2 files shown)


erichuanggit:
Generally It looks good to me. As discussed with ryang, he'd like to add this doc to the `Best Practise Book`

lubomir:
Nice job. Did you create the diagrams by hand? I'm wondering if for the data flow an image would be easier to read. For the second diagram it seems to me it is conflating two different kinds of information: required methods in PDC Messenger (`send_message` is called `send_msg` in the first diagram) and some details about existing implementations. I think the second part would be better presented in a table

```
Messenger | Dependencies | Notes
----------+--------------+-----
Dummy     | none         | ignores messages
```

I would like to see some guidance on how to choose a messenger. The dummy and fedmsg ones are obvious, but the other three look to be doing almost the same thing.

xychu:
@lubomir thanks for your comments.
I used http://asciiflow.com/, and diagrams can be saved into Google Drive.
And I agree that the second mixed two things together, I'll try to make it more clearly.
Also, after splitting notes about current implementations out, it should be easier to explain how to choose one messenger over others.

lubomir:
:+1:

PR #56, opened 2015-08-17
Add dud as valid content format

lubomir:
It stands for Driver Update Disk. Staging and production servers were
updated already, this is only a fix so that if we ever need to recreate
the database from scratch it will automatically created again.

JIRA: PDC-932


erichuanggit:
:+1: 
this content format need to be added to staging&prod by manual, as database won't be wiped. This migration file won't be applied.

lubomir:
Prod and stage servers are already updated.

PR #58, opened 2015-08-17
Create compose-images resource

lubomir:
This is a resource that works pretty much the same as compose-rpms. It
allows import compose images and getting the same data back.

The old import API is still available, but marked as deprecated.

Since the existing code did not store path for the images (just the
filename), there are two migrations to get to the desired state. The
first migration deletes all ComposeImage objects, the second creates the
new column in database.

The idea is that deploying on the server and running the migrations will
only lose the minimum amount of data. The images themselves will be
preserved, only their connections to composes will need to be imported
again.

JIRA: PDC-911

### DIFF
--- a/pdc/apps/compose/fixtures/tests/compose_composeimage.json
+++ b/pdc/apps/compose/fixtures/tests/compose_composeimage.json
@@ -1,10 +1,18 @@
 [
   {
+    "model": "compose.Path",
+    "pk": 1,
+    "fields": {
+        "path": "path/to/images"
+    }
+  },
+  {
     "model": "compose.ComposeImage",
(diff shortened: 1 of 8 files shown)


pdc/apps/compose/migrations/0004_auto_20150817_1038.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
erichuanggit: missing license, same for the other migration file.

erichuanggit:
Generally looks good to me.

erichuanggit:
:+1:

PR #59, opened 2015-08-17
Provide list of filters automatically

lubomir:
More details are [in the commit message of the first commit](https://github.com/release-engineering/product-definition-center/commit/a67676235390ada6c9f13f932ec9ed1a5d712fe6).


erichuanggit:
:+1: 
I like this documentation automatically

lubomir:
I merged this so that it does not block further work on doc automation. We can modify it later anyway.

PR #60, opened 2015-08-18
Catch more detail excepton.

erichuanggit:
Catch ValueError exception for release component input.
It's not friendly to tell user internal error but log into
server's error log file without this exception block.

JIRA: PDC-905


lubomir:
This is ok, but it seems to me that adding `convert_str_to_int` to line 452 would be easier and produce nicer error message.

erichuanggit:
I was assuming there will be other fields except for id causing exception since we support id or unique_together fields input. 
`'{"release": 1, "group_type": 22,  "description": "dd",  "components":[{"release": "rhel-7.0", "global_component": "Cython", "name": "Cython"}]}'`
As a quick testing, looks like the other fields have the validation.

simozhan:
Ok for me

lubomir:
:+1:

PR #62, opened 2015-08-18
Auto urls

lubomir:
This set of patches replaces hardcoded urls in documentation with automatically generated ones. Most changes are in the renderer class, the rest is pretty much just replacing existing links (there's a lot of them).

I'm not sure the test is a good idea as it is quite slow.

### DIFF
--- a/pdc/apps/auth/views.py
+++ b/pdc/apps/auth/views.py
@@ -94,7 +94,7 @@ class TokenViewSet(StrictQueryParamMixin, viewsets.ViewSet):
 
     * obtain token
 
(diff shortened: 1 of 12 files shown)


erichuanggit:
if it is really slow, how about add a flag in settings_test.py to control whether to run it? At the meantime, we could dig into to see the root cause.

erichuanggit:
:+1: 
It only takes 1.5 second for this single test running in my local. I think it is worth to have it.

lubomir:
In my local it also takes about 1.5 seconds, while the whole test suite needs about 25 secs. I would also prefer keeping this, as for development it is often more convenient to run just part of the tests with `python manage.py test pdc.apps.XXX`.

simozhan:
Cool. +1

PR #63, opened 2015-08-19
Display all routes in API root

lubomir:
This patch modifies the router to include all resources in the API root.
It tries various options how to generate an URL. Currently it works with
all existing resources. If in the future we include a resource it can
not generate a url for, the resource will be listed with null url.

There is a bit of a heuristic to avoid generating urls that will lead to
a server crash.

A test is included that tests more special urls.

JIRA: PDC-913


erichuanggit:
Looks good to me. But need to rebase and resolve conflict.

PR #65, opened 2015-08-20
Add links to existing resources.

erichuanggit:
JIRA: PDC-916


lubomir:
Looks good to me. I would remove the Query Params header from documentation since there aren't any.

erichuanggit:
Fixed in second commit. Thanks

simozhan:
+1

PR #67, opened 2015-08-21
Split compose migrations in two

lubomir:
PostgreSQL does not like a migration to modify data and change the
schema at the same time, so use first migration to fill in missing paths
for compose images and second migration to make the field non-nullable.

### DIFF
--- a/pdc/apps/compose/migrations/0005_auto_20150819_0827.py
+++ b/pdc/apps/compose/migrations/0005_auto_20150819_0827.py
@@ -24,9 +24,4 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunPython(set_empty_path),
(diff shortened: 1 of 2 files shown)


lubomir:
To replicate, check out commit 16b2f00, create database with some compose image instance (sync.sh should be enough), check out master, try to apply migrations. This fails. Try with this commit instead of master, it should work.

erichuanggit:
:+1:
met the same issue for release_component_type in dev and qa server. But can't reproduce it on my local environment. 
But I would suggest we do the same thing with this patch.

xychu:
:+1:

simozhan:
Cool. +1. Just met this issue in QA server and fixed it in the same way.

PR #68, opened 2015-08-21
Add migration to set rpm as type to all components

lubomir:
This needs to be done before component type is made non-optional and
prevents migration error on PostgreSQL 9.2.


erichuanggit:
Looks good to me.

simozhan:
+1

bliuredhat:
+1

PR #69, opened 2015-08-21
Auto serializer

lubomir:
Please see commit messages on the first two commits to get more details.

There are some code changes in renderer, other than that it is mostly documentation adjustments.


erichuanggit:
Looks good to me.
I like this change. It will reduce lots of manual doc for any new endpoint in future. Good job lubomir!!

simozhan:
Yeah. Looks pretty good to me. Data and Response sorts in order with meaningful description.
PDC-890 and its sub tasks will be updated after this patch merges.

bliuredhat:
+1

PR #71, opened 2015-08-24
Add base_name for compose-images and build-images

sochotnicky:
This way they can be referenced easily

### DIFF
--- a/pdc/routers.py
+++ b/pdc/routers.py
@@ -74,7 +74,8 @@ router.register(r'composes/(?P<compose_id>[^/]+)/rpm-mapping',
                 compose_views.ComposeRPMMappingView,
                 base_name='composerpmmapping')
 router.register(r'compose-rpms', compose_views.ComposeRPMView)
-router.register(r'compose-images', compose_views.ComposeImageView)
+router.register(r'compose-images', compose_views.ComposeImageView,
+                base_name='composeimages')
 
 router.register(r'rpc/release/import-from-composeinfo',
                 release_views.ReleaseImportView,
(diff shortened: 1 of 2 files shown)


pdc/routers.py
```diff
@@ -74,7 +74,8 @@
                 compose_views.ComposeRPMMappingView,
                 base_name='composerpmmapping')
 router.register(r'compose-rpms', compose_views.ComposeRPMView)
-router.register(r'compose-images', compose_views.ComposeImageView)
+router.register(r'compose-images', compose_views.ComposeImageView,
+                base_name='composeimages')
```
erichuanggit: need to update base_name to `composeimage`, same as BuildImage. Otherwise related unit tests can't be pass.

pdc/settings.py
```diff
@@ -29,7 +29,7 @@
 SECRET_KEY = '3hm)=^*sowhxr%m)%_u3mk+!ncy=c)147xbevej%l_lcdogu#+'
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = False
+DEBUG = True
```
erichuanggit: I don't think we need to make this change in settings.py. Generally we will override it in settings_local.py.

sochotnicky: Oh yeah, I was actually sure I didn't include this change :/ Sorry

lubomir:
This is not necessary. The base name only needs to be manually specified if the viewset does not use a model. In this case, there alredy is a basename of `composeimage`, which can be referenced by `composeimage-list` and `composeimage-detail`. Please note that unit tests already use these names.

sochotnicky:
Meh, I realized this is not needed. There was a typo in my template

PR #72, opened 2015-08-24
Make sigkey key_id editable

lubomir:
This is consistent with other resources where their string primary
identifier can be changed as well (release_id, product_version_id etc).

This change makes it possible to allow PUT requests to update keys. The
original reason to disable it was because key_id was read-only.

JIRA: PDC-634, PDC-855

### DIFF
--- a/pdc/apps/common/tests.py
+++ b/pdc/apps/common/tests.py
@@ -337,8 +337,11 @@ class SigKeyRESTTestCase(TestCaseWithChangeSetMixin, APITestCase):
 
     def test_update_sigkey(self):
         url = reverse('sigkey-detail', kwargs={'key_id': '1234adbf'})
-        response = self.client.put(url, format='json', data={'name': "TEST"})
-        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
+        response = self.client.put(url, format='json',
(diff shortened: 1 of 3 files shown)


pdc/apps/common/tests.py
```diff
@@ -346,21 +349,26 @@ def test_partial_update_sigkey(self):
...
-        self.assertEqual(response.data.get('key_id'), '1234adbf')
-        self.assertNumChanges([])
+        self.assertEqual(response.data.get('key_id'), 'cafebabe')
+        self.assertNumChanges([1])
+        response = self.client.get(reverse('sigkey-detail', args=['1234adbf']))
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+        response = self.client.get(reverse('sigkey-detail', args=['cafebabe']))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
 
     def test_bulk_update_fails(self):
```
erichuanggit: It's better change this test case's name.

simozhan:
As key_id is not read_only, maybe we can change __Data__ from %(SERIALIZER)s to %(WRITABLE_SERIALIZER)s in SigKeyViewSet.create to make it consistent with other APIs.

erichuanggit:
Looks good to me.

simozhan:
Good to me.

PR #74, opened 2015-08-25
Make doc of global comp. labels consistent

lubomir:
The documentation used different format than all other resource.

The create action documentation now does not mention the description
field which is not used anyway. See PDC-611.

The links in documentation lead to page causing internal server error.
While unfortunate, this already happens on global/release component
contacts


simozhan:
+1. Maybe give a warning to tell the user to input int instead of {component_id} or something?

erichuanggit:
Looks good to me.

lubomir:
@simozhan Good point, I think I actually fix the internal server error.

simozhan:
Looks good to me

PR #78, opened 2015-08-26
Repo filters should accept duplicit filters

lubomir:
Since we expect to have a lot of repos, the filtering should allow
queries with multiple values to choose from (OR). Without this patch,
multiple occurrences of the same filter are accepted, but only the last
value is used for filtering.

JIRA: PDC-970

### DIFF
--- a/pdc/apps/repository/filters.py
+++ b/pdc/apps/repository/filters.py
@@ -10,13 +10,13 @@ from . import models
 
 
 class RepoFilter(filters.FilterSet):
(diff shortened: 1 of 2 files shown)


erichuanggit:
:+1: 
It will be nice if have the unit test cover this case. But it's not a big deal. I believe we already had the unit test to cover `MultiValueFilter`.

lubomir:
@erichuanggit The tests are definitely a good idea, if only to prevent future regressions.

I will squash the commits before merging.

PR #80, opened 2015-08-27
Separate auto doc test

lubomir:
Move test for auto documentation to a separate module that is not
included in make test. The reason for this is that the test hits most
end-points, but does not check much about them. This is raising the test
coverage significantly which could hide bugs.

The test is still run on Travis-CI and can be run locally via
    $ make extra_test

JIRA: PDC-975

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
 script:
     - coverage run --source=contrib,pdc --omit='*tests.py,*/migrations/*,pdc/settings*' manage.py test --settings pdc.settings_test
     - make flake8
+    - make extra_test
 after_success:
     coveralls
 cache:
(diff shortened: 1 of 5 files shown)


erichuanggit:
:+1: 
I agree this change by de-coupling auto doc test case

bliuredhat:
+1

simozhan:
Cool. I like this change to move test for auto documentation to a separate module.
Maybe adding a test to only list api-root in pdc/apps/common/tests.py can increase the coverage a bit. : ). Minor thing.
+1

lubomir:
@simozhan The point of this patch is for the coverage to reflect which code is actually tested, not just executed with no assertions. I think there should be separate tests for the renderer itself.

simozhan:
@lubomir Agree. There should be seperate tests for renderer, but that's not related to this fix.

PR #82, opened 2015-08-27
Use automatic serializer for component types

lubomir:
This patch was merged at about the same time as the automatic serializer
support and it was not used yet.

JIRA: PDC-977


erichuanggit:
Looks good to me.

PR #83, opened 2015-08-31
Improve web view of repos

lubomir:
This patch adds total count of repos, displays their primary ids
(useful for API manipulation).

It also enables sorting and filtering of the table.

JIRA: PDC-978


bliuredhat:
+1

PR #84, opened 2015-08-31
Report error when update specifies a read-only key

lubomir:
The fields would otherwise be ignored, which is confusing for the user.
This change required fixing a couple tests. There were tests that
actually checked that the fields are ignored, now they verify that the
request fails.

A bigger change was needed in repository clone view to make sure ids are
not fed into the serializer. This actually fixes an error when
previously the response of repo clone would include old identifiers.

JIRA: PDC-951

### DIFF
--- a/contrib/drf_introspection/serializers.py
+++ b/contrib/drf_introspection/serializers.py
@@ -11,6 +11,15 @@ from rest_framework import serializers
 from django.core.exceptions import FieldError
 
 
+def _error_with_fields(message, fields):
+    """
+    Helper function to create an error message with a list of quoted, comma
+    separated field names. The `message` argument should be a format string
+    with a single `%s` placeholder.
+    """
+    return message % ', '.join('"%s"' % f for f in fields)
+
+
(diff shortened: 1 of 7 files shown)


contrib/drf_introspection/serializers.py
```diff
@@ -70,4 +81,19 @@ def maybe_raise_error(extra_fields):
         serializer, but still wants to validate its input.
         """
         if extra_fields:
-            raise FieldError('Unknown fields: %s.' % ', '.join('"%s"' % f for f in extra_fields))
+            raise FieldError(_error_with_fields('Unknown fields: %s.', extra_fields))
+
+    def check_read_only_fields(self, keys):
+        """
+        Check that all fields are not read-only. If some are, a ValidationError
```
simozhan: ValidationError should be changed to FieldError. i guess this is a typo.

bliuredhat:
+1

erichuanggit:
Looks good to me.

PR #89, opened 2015-09-01
Experiment for with different approach to contact manipulation

ycheng-aa:
Support endpoints /global-component-contacts/ and
/release-component-contacts/.


pdc/apps/component/serializers.py
```diff
@@ -534,3 +535,66 @@ class ReleaseComponentRelationshipSerializer(StrictSerializerMixin, serializers.
...
+    role = serializers.SlugRelatedField(source='contact_role', slug_field='name',
+                                        read_only=False, queryset=ContactRole.objects.all())
+    contact = ContactField()
+
+    class Meta:
+        model = RoleContact
+        fields = ('id', 'component', 'role', 'contact')
+
+
+class ReleaseRelatedField(ReleaseComponentRelatedField):
```
erichuanggit: What's field for? the inheritance looks strange from the class names.

ycheng-aa: This is useless. I forgot remove it. Thanks for finding it.

lubomir:
The output has the format that looks very usable to me. Unfortunately, it does not seem to work correctly for inheritance of release component contacts.

I tried with almost empty database, with a single global component and single release component. Define a contact for the global component – bob@example.com as Build_Owner. Then define a contact for the release component – alice@example.com as QE_Owner. Visit `/release-component-contacts/` API. Only the contact from release component is included, while there should be both (one defined directly on release component, the other inherited from global component).

If I delete the contact from release component, the release component contact API will start showing the inherited one.

It looks like the contacts are inherited correctly only if the release component has no contacts on its own.

ycheng-aa:
@lubomir Yes. It is what I understood the requirement. I thought if release component has any contact then use its contact information and don't care the global component's information. If rc don't have contact information at all, then use global component.

So the correct understanding should be if global component has any contact role information that rc don't have, this contact role information should be corresponding rcs' contact role info, right?

lubomir:
The inheritance should be the same as is now: release component takes all contacts from global component, and overwrites it with anything it has defined on its own. So your second paragraph is correct. I'm sorry if the original task description was unclear.

ycheng-aa:
@lubomir Thanks. I tried to implement as you said in 14bf5c556f54fdd7. But I changed the model structure back. Because I found in original model structure, it looks more convenient in cloning release, changing contact at one place for all global components and release components as the task mentioned, etc.

lubomir:
I'm not sure how this is supposed to be used. The original proposal included identifiers for each component contact that were unique. In this version when I visit `/global-component-contacts/`, I get a page that contains contacts with identifiers 1 to 20, but for each id there are multiple contacts. How is the user supposed to pick one which they want to edit?

For release components it seems to work the same.

I tried to put some examples in PDC-912 to hopefully better illustrate the original intention. This is not a hard requirement if something better is found.

ycheng-aa:
@lubomir Thanks for the review and update.

I think the output should not like you said " for each id there are multiple contacts.". Each id's contact should be exactly same. This id is role contact's ID, the contact and role should be same for one id. The difference is component (and component-id for release component). Fro release component, the component should be release component that is connected to this contact role directly and the release component which don't have this role's contact directly but its global component is connected to this role contact.

For the question to edit contact, if user change a role contact's contact. Then it take effect to all global component and release component that are connected to this role contact. For example, curl -H 'Content-Type: application/json' "http://127.0.0.1:8000/rest_api/v1/release-component-contacts/35/" -X PATCH -d '{"contact": {"mail_name": "IdM - Red Hat Directory Server"}}'.

The solution you suggested is each release component and each global component its own contact and role, even some of them have same contact and role, but they don't share contact and role as current. One record in db table is (contac_id, role id, release component/global component id), NOT (role contact id, release component/global component id) any more. In your example global-component-contacts/1234, the 1234 is id of record (role contact id, global component id). 

When we change role or contact information, it will only affect this specific release/global component, not affect other components. Currently if you change contact or role in one role contact record, it will affect all related components. In your suggestion, if we want to replace contact at one place, we will change Contact model content directly rather than RoleContent. Is the understanding correct? Sorry to bother again. Thank you very much for the patience and help.

ycheng-aa:
@lubomir I think again, maybe I understood above is incorrect. It looks difficult to show inherited information. If we separate global component contact in table with record item being (role id, contact id, gc id) and release component contact with record item (role id, contact id, rc id), It is not a property id to show inherited rc contact when list rc contacts. Because it is from gc contacts table. Is it possible to keep old model structure and manipulate role contact, and global components and release components are all connected to role contact?

PR #90, opened 2015-09-01
Enable storing RPM dependencies

lubomir:
Each RPM can have arbitrary number of dependencies of multiple types.
The types are hard-coded, and can not be extended at run-time.

With this patch, it is possible to store and retrieve the dependencies.
There are tests for these parts.

Existing tests have been updated to pass: for RPM tests there are some
minor updates. The compose/package API and its variants are updated not
to include dependencies in their output.

All changes to dependencies are logged in change sets as single update
to the RPM.

The filtering support is not finished yet. So far it is only possible to
get RPMs that depend on a given package with optional equality version
constraint. This is not documented nor covered with unit tests.

JIRA: PDC-955

* [x] Implement filters.
* [x] Check that filters work correctly for all versioning schemes.
* [x] Support versions with epoch.
* [x] Squash migrations before final merge.

### DIFF
--- a/pdc/apps/package/serializers.py
+++ b/pdc/apps/package/serializers.py
@@ -9,7 +9,7 @@ from django.contrib.contenttypes.models import ContentType
 from rest_framework import serializers
 
 from . import models
-from pdc.apps.common.serializers import StrictSerializerMixin
+from pdc.apps.common.serializers import StrictSerializerMixin, DynamicFieldsSerializerMixin
 
 
 class DefaultFilenameGenerator(object):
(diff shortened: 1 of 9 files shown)


pdc/apps/package/serializers.py
```diff
@@ -46,6 +79,24 @@ def to_internal_value(self, data):
             data['filename'] = models.RPM.default_filename(data)
         return super(RPMSerializer, self).to_internal_value(data)
 
+    def create(self, validated_data):
+        dependencies = validated_data.pop('dependencies', [])
```
erichuanggit: Default value for `dependencies` field is `{}`. Shall we make it consist here?

lubomir: I don't think so. The `DependencySerializer` field takes dict on input, but after validating it returns just a list, as the split into 6 different lists is no longer needed. I guess this should be better documented for developers.

lubomir:
The patches are ready for review. I would squash them before merging, but I'm not sure if the refactor is actually an improvement – the code is shorter, but I'm not sure it actually is easier to read.

xychu:
LGTM, really great work on the tests.

PR #91, opened 2015-09-02
Move out a single endpoint from "compoese/package"

erichuanggit:
Create a new endpoint called
"rpc/find-composes-by-product-version-rpm/{product_version}/{rpm_name}".
The old endpoint "compose/package" has been marked to deperated.
Unit test also added.

JIRA: PDC-944

### DIFF
--- a/pdc/apps/compose/tests.py
+++ b/pdc/apps/compose/tests.py
@@ -52,6 +52,7 @@ class ComposeModelTestCase(TestCase):
 
 
 class VersionFinderTestCase(APITestCase):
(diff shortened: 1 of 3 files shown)


bliuredhat:
+1

lubomir:
Looks good to me. Since all functionality of the old ViewSet is now available elsewhere, I would remove its documentation so that only the links to new api are there. This should discourage use of the deprecated api a bit.

erichuanggit:
Right. the document will be removed. I will rebase with master and push it directly.

PR #92, opened 2015-09-02
PDC-947,the boolean values for filters should not be case sensitive

bliuredhat:
Make bootable filter on images case-insensitive

JIRA: PDC-947

### DIFF
--- a/pdc/apps/common/filters.py
+++ b/pdc/apps/common/filters.py
@@ -214,7 +214,7 @@ class NullableCharFilter(django_filters.CharFilter):
         return super(NullableCharFilter, self).filter(qs, value)
 
 
-class CustomizeBooleanFilter(django_filters.CharFilter):
+class CaseInsensitiveBooleanFilter(django_filters.CharFilter):
     """
     Customize boolean filter that allows filtering items with below values:
     True: ('true', 't', '1'), regardless of letter case sensitive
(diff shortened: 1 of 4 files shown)


lubomir:
Generally looks good. I would like the filter to actually be called something like `CaseInsensitiveBooleanFilter`, but that is likely out of scope for this issue. One minor thing: in order to have nice git history, please change the commit message a bit – put the issue number in the body. I would also suggest different wording:
```
Make bootable filter on images case-insensitive

JIRA: PDC-947
```
These two links have some guidelines about good commit messages. These are not hard rules, just something to think about:
http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
http://chris.beams.io/posts/git-commit/

bliuredhat:
lubomir, thank you so much~! The link you surported is very good for me.  I will rewrite the commit message.

bliuredhat:
lubomir, The 'CustomizeBooleanFilter' function in [pdc/apps/package/filters.py] is not defined by me, so I can't redefine the name.

erichuanggit:
Looks good to me.

requirements/devel.txt
```diff
@@ -10,6 +10,6 @@ setuptools
 lxml
 # model graph
 # requires: graphviz-devel
-pygraphviz
+pygraphviz<1.3
```
erichuanggit: Remove this change, since it has been fixed in pygraphviz 1.3.1

PR #94, opened 2015-09-03
Fixes for default filename

lubomir:
There are two mostly unrelated patches in this PR. Both come from the same issue. First patch changes the format of error message of bulk update. The second patch fixes how default RPM filename is generated. See commit message of the second patch for more details.

JIRA: PDC-959


erichuanggit:
Looks good to me.

simozhan:
Good to me.

PR #95, opened 2015-09-03
Report error when importing inconsistent data

lubomir:
When images or rpms are imported and compose id in composeinfo and
rpm/image manifest is not the same, report an error. This could happen
when user picks manifest for different compose by mistake.

It does not do any sophisticated validations nor does it compare any
other fields.

JIRA: PDC-952


erichuanggit:
Looks good to me.

PR #96, opened 2015-09-04
Fix RPM epoch filter to not crash on non-numeric values

lubomir:
This patch also fixes the documentation to properly show the value
should be an int.

JIRA: PDC-997


erichuanggit:
Looks good to me.

PR #97, opened 2015-09-06
jira PDC-960, Order resources in client

bliuredhat:
Print resources in order when running  'pdc_client -s qa ' shell command.

jira: PDC_960

### DIFF
--- a/pdc_client/bin/pdc_client
+++ b/pdc_client/bin/pdc_client
@@ -182,4 +182,4 @@ if __name__ == "__main__":
         with RequestMethod(client, session, options.resource,
                            options.traceback, options.debug) as request:
             print json.dumps(request.dispatch(options.request, data),
-                             indent=4)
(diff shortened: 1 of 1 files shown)


requirements/devel.txt
```diff
@@ -10,6 +10,6 @@ setuptools
 lxml
 # model graph
 # requires: graphviz-devel
-pygraphviz
+pygraphviz<1.3
```
erichuanggit: I would suggest having a single patch to fix it, but not include this change in every patch.

erichuanggit:
Generally looks good to me.

simozhan:
Good to me

requirements/devel.txt
bliuredhat: erichuanggit, yes， I will do；Thank you！

lubomir:
Looks good to me. I think the commit message would deserve a bit more love.

PR #98, opened 2015-09-06
Add email notification as meet any error

erichuanggit:
This feature is disable by default. To enable it:
* cope settings_local.py.dist and rename it to settings_local.py
* make sure DEBUG set to False
* un-comment content as what describes at the end of example file

JIRA: PDC-991

### DIFF
--- a/pdc/apps/common/handlers.py
+++ b/pdc/apps/common/handlers.py
@@ -58,8 +58,6 @@ def exception_handler(exc, context):
         else:
             logger = logging.getLogger(__name__)
             logger.error('Unhandled exception', exc_info=sys.exc_info())
-            return Response({'detail': 'The server encountered an internal '
-                                       'error or misconfiguration and was '
-                                       'unable to complete your request.'},
+            return Response(data=settings.INTERNAL_SERVER_ERROR_RESPONSE,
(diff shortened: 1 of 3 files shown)


erichuanggit:
The error causes by:
https://github.com/pygraphviz/pygraphviz/issues/69

erichuanggit:
It has been fixed in pygraphviz 1.3.1

bliuredhat:
+1

erichuanggit:
Fixed in second patch. the real email configuration will be put into deploy guide.

erichuanggit:
just realize it is not correct to add eng-ops@redhat.com, since this change is in public code.
comment from weekly pad: add more info into the 500 error response; e.g. file ticket with eng-ops@redhat.com

lubomir:
@erichuanggit How about moving the text to settings? There already is `EMPTY_PATCH_ERROR_RESPONSE`, so we could add `INTERNAL_SERVER_ERROR_RESPONSE`. In upstream, we could use the current text, and in internal version add the contact details.

erichuanggit:
no more better choice. thanks for suggestion.

lubomir:
:+1:

xychu:
:+1:

PR #99, opened 2015-09-07
Keep track of reasons for changes.

simozhan:
As a user I want to keep track of reasons for changes

JIRA: PDC-923

### DIFF
--- a/pdc/apps/changeset/filters.py
+++ b/pdc/apps/changeset/filters.py
@@ -23,6 +23,7 @@ class ChangesetFilterSet(django_filters.FilterSet):
                                                 widget=widgets.DateTimeInput)
     changed_until = django_filters.MethodFilter(action='filter_committed_until',
                                                 widget=widgets.DateTimeInput)
+    comment = django_filters.CharFilter(name="comment", lookup_type="contains")
 
     @value_is_not_empty
     def filter_author(self, qs, value):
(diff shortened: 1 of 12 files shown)


simozhan:
I changed the existing middleware test cases which made me feel not very good. Any good suggestion?

erichuanggit:
is this supposed on `bulk`? if yes, could we add related test cases?

simozhan:
I am assuming if the change is ok for create and bulk create. It should be ok for the other operations. right?

lubomir:
This looks ok to me. I think it would be better to rename to field to {{change_comment}} or maybe just {{comment}}. There is no reason for the {{pdc_}} prefix. The reason why it is in the header name is so that it is clear that part of the HTTP request is PDC specific. But in database, everything is PDC related, so we don't need to explicitly prefix columns with that info.

simozhan:
Yes. pdc prefix is not necessary. Fixing

lubomir:
I found one more issue and one possible improvement:
 * the example for changeset list/retrieve API needs to be updated to display the field (response format section is correct)
 * it would be nice to add the comment to `pdc/apps/changeset/templates/changeset_detail.html`

simozhan:
Thanks @lubomir for pointing out this issue. Totally forgot it. Fixed.

lubomir:
:+1: Please squash the commits before merging.

PR #100, opened 2015-09-07
Update doc for end-point rpc/where-to-file-bugs

erichuanggit:
JIRA: PDC-976


lubomir:
I think the example should either be removed or show some actual example data. Maybe product *Fedora* and component *GConf2* would work well.

erichuanggit:
Fixed.

lubomir:
:+1:

PR #101, opened 2015-09-07
Avoid losing empty list with pdc_client

erichuanggit:
With pdc_client, not only for empty list, but empty dictionary,
it will be lost. Convert it to empty string to avoid this case.

Not sure if it is the right fixing.

JIRA: PDC-948

### DIFF
--- a/pdc_client/bin/pdc_client
+++ b/pdc_client/bin/pdc_client
@@ -170,7 +170,7 @@ if __name__ == "__main__":
     data = load_data(options)
 
     for key, value in data.iteritems():
-        if not value:
+        if options.request == 'GET' and not value:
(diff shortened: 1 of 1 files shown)


lubomir:
Do we actually want to fix this? I could reason that this is not a bug. Generally the best fix for this would be to provide interface where users do not have to write JSON to specify query filters. We already have an issue open about providing user-friendly interface.

I think the proposed fix could work, although I didn't test it. There is an issue with it: it should only apply to query filters, never to request body.

erichuanggit:
To be honest. I don't like this change either. It's not worth to hack for an useless case. I will leave it here for a while and get more feedback from team.

PR #102, opened 2015-09-07
Create role contact with non-existing role type is not allowed.

ycheng-aa:
JIRA: PDC-982

### DIFF
--- a/pdc/apps/component/tests.py
+++ b/pdc/apps/component/tests.py
@@ -393,7 +393,7 @@ class GlobalComponentContactRESTTestCase(TestCaseWithChangeSetMixin, APITestCase
         # PDC-260
         url = reverse('globalcomponentcontact-list', kwargs={'instance_pk': 2})
         data = {
-            "contact_role": "new_type",
+            "contact_role": "qe_ack",
             "contact": {
                 "username": "person3",
                 "email": "person3@test.com"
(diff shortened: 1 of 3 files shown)


lubomir:
I think it is still possible to to replicate the problem via component contacts:
```
pdc_client -s local -r release-components/1/contacts -x post -d '{"contact_role": "Buld_Owner", "contact": {…}}'
```

This should be fixed too.

erichuanggit:
Other fixing should be in class `RoleContactSpecificManager` of contact.models

erichuanggit:
@lubomir, I am not sure if we need to prevent this nested resource created. At lease it is not consistent. There are still some cases need to be improved as well. List some of them but not limited:
> Create a RoleContact with non existed Person `username` or Maillist `mail_name`, Person/Mailist will be created.

lubomir:
You're right, the contact will also be created if it did not exist before. Maybe @sochotnicky can tell us what the correct behaviour should be.

ycheng-aa:
Fixed in 4438e72f2702ec274e8631ee29bd. I kept the release-components/1/contacts behavior no touch on purpose. Now it should prevent this kind of indirect creation. But only for contact role now.

As discussed with @erichuanggit, if only prevent new contact role from being created, RoleContactSpecificManager don't have to be changed.

simozhan:
I like the last commit. Only prevent new contact role from being created. +1

lubomir:
I like it too. I never meant for person or maillist to not be created, just the contact role. Could the `ContactRoleField` be replaced with `ChoiceSlugField`? It would improve the error message.

ycheng-aa:
Changed it to ChoiceSlugField. Then ContactRoleField became useless. Also removed it.

lubomir:
:+1:

erichuanggit:
Looks good to me.

PR #103, opened 2015-09-08
Track request duration in changesets.

simozhan:
JIRA: PDC-740

### DIFF
--- a/pdc/apps/changeset/migrations/0004_changeset_requested_on.py
+++ b/pdc/apps/changeset/migrations/0004_changeset_requested_on.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('changeset', '0003_changeset_pdc_change_comment'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='changeset',
+            name='requested_on',
+            field=models.DateTimeField(null=True),
+        ),
+    ]
--- a/pdc/apps/changeset/views.py
+++ b/pdc/apps/changeset/views.py
@@ -68,6 +68,8 @@ class ChangesetViewSet(StrictQueryParamMixin,
 
         %(SERIALIZER)s
 
+        The unit for duration is second.
+
         __Example__:
 
             curl -H "Content-Type: application/json"  -X GET $URL:changeset-list$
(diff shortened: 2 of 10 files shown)


lubomir:
Looks good to me.

pdc/apps/changeset/views.py
```diff
@@ -80,7 +80,9 @@ def list(self, request, *args, **kwargs):
                     {
                         {
                             "author": "xxx",
+                            "requested_on": "2015-02-03T05:50:49.387Z",
                             "committed_on": "2015-02-03T05:51:17.262Z",
+                            "duration": "27.875",
```
erichuanggit: Is it better to append `unit`?

lubomir: I don't think so. At most the unit could be mentioned in documentation, but putting it in the response does not seem very useful.

erichuanggit:
yes, what i want to mentioned is adding to doc.

lubomir:
:+1:

pdc/apps/changeset/migrations/0004_changeset_requested_on.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('changeset', '0003_changeset_pdc_change_comment'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='changeset',
+            name='requested_on',
+            field=models.DateTimeField(default=datetime.datetime(2015, 9, 8, 2, 15, 31, 582918, tzinfo=utc)),
```
xychu: Will this set existed records 'requested_on' to '2015-9-8'?
What will happen if theirs 'committed_on' are before that value?
I think 'requested_on' should allows to be Null and we could set 'duration' to NaN for historical data that without 'requested_on'.

lubomir: Another option is to set `requested_on` to the same value as `commited_on`, and old data would show as taking 0 seconds.

xychu: Much simpler and neater. :+1:

ycheng-aa:
Looks good. And I agree with @erichuanggit , add time unit to doc may be more clear.

pdc/apps/changeset/migrations/0004_changeset_requested_on.py
simozhan: ok. will fix it

xychu:
I think @lubomir means we could add a data migration that set the historical changesets' requested_on to the value of theirs committed_on. So it's more like:
* a migration to add field that allow null;
* a data migration to update historical data;
* another migration to update field that no allow null;

lubomir:
Yes, @xychu is right. Doing it with these migrations means there would be no hacks in code, and the field would be required, which is also desirable.

lubomir:
Looks good to me. Please squash before merging.

xychu:
:+1:

PR #104, opened 2015-09-08
Disable paging of repos in release detail

lubomir:
The paging on that page is not useful in any way. It was previously
disabled through data attributes, but since test server is using an
outdated version of PatternFly and therefore DataTables, this patch
explicitly disables the paging with JavaScript.

JIRA: PDC-978


erichuanggit:
Looks good to me.

PR #105, opened 2015-09-08
Fix empty response docs

lubomir:
A couple end-points provided incorrect documentation for response when
deleting. The server does not return anything (on success). It is the
client that prints the "No content" message.

JIRA: PDC-980


erichuanggit:
Looks good to me.

PR #107, opened 2015-09-08
When bulk destroy, response include detailed removed items information.

ycheng-aa:


lubomir:
I don't think this is the right way to fix this. If delete is supposed to return deleted object, then it should be done on a regular delete of a single object. Bulk operations would be modified to use response from that.

My reasoning is that I don't see any reason why bulk delete should return the deleted content but single delete should not. When user does delete on `/resource/1/`, it should work exactly the same as delete on `/resource/` with data `[1]`.

Personally, I would say this is not a bug. If the user knows the id, they know what they are deleting, and they can get the details from the changeset anyway.

If this patch is merged, the documentation needs to be updated.

ycheng-aa:
@sochotnicky What is your opinion? Should we change all the 'destroy' response and corresponding doc? The task is PDC-940.

PR #108, opened 2015-09-09
"not found" enhancement for buld destory overrides with query param.

ycheng-aa:
Return 404 instead of 200 if not found any matched objects.
JIRA: PDC-940


lubomir:
:+1:

erichuanggit:
Looks good to me.

PR #110, opened 2015-09-10
Add pdc-client spec file.

xychu:
Move client docs under 'pdc_client/'.

Test `python-pdc`:
* (at the project root dir, which is same level of `pdc.spec`)
* `tito tag` or `tito tag --keep-version`
* `tito build --rpm --test`
* `tito release copr-pdc-test --dry-run`

Test `pdc-client`:
* (change directory to the same dir level of `pdc-client.spec`)
* other commands are same as above


lubomir:
Looks good to me.

erichuanggit:
Looks good to me.

PR #111, opened 2015-09-10
Not convert *False* filter.

erichuanggit:
Boolean *False* is possible to be a filter of some resource.
Need to exclude it in this hack.

### DIFF
--- a/pdc_client/bin/pdc_client
+++ b/pdc_client/bin/pdc_client
@@ -173,7 +173,7 @@ if __name__ == "__main__":
     data = load_data(options)
 
     for key, value in data.iteritems():
-        if options.request == 'GET' and not value:
+        if options.request == 'GET' and not value and value is not False:
             # empty string won't be lost
             data[key] = ''
 


pdc_client/bin/pdc_client
```diff
@@ -173,7 +173,7 @@ if __name__ == "__main__":
     data = load_data(options)
 
     for key, value in data.iteritems():
-        if options.request == 'GET' and not value:
+        if options.request == 'GET' and not value and value.lower() is not 'false':
```
ycheng-aa: I think it may cause exception if value is a empty list or dict. Maybe last condition just "value is not False" or "value != False" should be fine. String "false" will not pass 'not value' check. It should not be the condition need to be ruled out.

erichuanggit: you are right. we don't need to care about the other False boolean format.

ycheng-aa:
:+1:

lubomir:
What if the hack checked directly for empty list and empty dict? It is still two conditions and should not match any false positives.

erichuanggit:
Not only `empty list` and `empty dict`, but `None` is another one will be lost. Not sure if there are others. That's why this hack not check directly for empty list and empty dict.

lubomir:
Ok then. No further comments from me.

PR #113, opened 2015-09-10
Include RPM filename in changeset

lubomir:
The changesets were missing this field. There is still a field
`linked_composes` that is not included in the changeset. This is because
the field can not change through any API.

JIRA: PDC-995


erichuanggit:
Looks good to me.

ycheng-aa:
Looks good.

PR #114, opened 2015-09-10
Promote cloning instead of creating new releases

lubomir:
Cloning has some advantages over creating a new release from scratch.
The documentation for create is updated to link to the clone API.

Some information was missing in the clone doc, it is added now.

JIRA: PDC-355


erichuanggit:
Looks good to me.

ycheng-aa:
Looks good.

PR #115, opened 2015-09-10
Two tito tags that indicate tito has entered the field.

xychu:
Two tags that add by `tito tag --keep-version`.

After that our changes will be managed by tito, so next time we want to have a release,
we just need to `tito tag`, it will help us to modify spec file, and include all changes ever since into the changelog.


lubomir:
Overall, this looks pretty good to me (although I would like it if the commit messages were shorter).

I just noticed one issue: in dev server, the version shown in footer is determined from `git describe`. Running that command on this branch, the result is *pdc-client-0.1.0-1*, which does not work with rest of code for splitting out the version. The fix would be to run `git describe --match=python-pdc-*` in `pdc/__init__.py`.

xychu:
Yeah, the glob pattern used by `git describe` after we have a name prefixed tag is straightforward.
In my local I hold a commit that support both 'vX.Y.Z*' and 'python-pdc*', but it is as ugly as
`-match='[p,v][y,0-9]*'`.
I will add one commit use the readable pattern as you suggested.

PR #116, opened 2015-09-11
Fix displayed value of RPM filename default

lubomir:
JIRA: PDC-1010


erichuanggit:
Looks good to me.

simozhan:
Looks good to me.

PR #117, opened 2015-09-11
Show current tab in URL

lubomir:
This way it is possible to link to a specific part of the documentation,
not just the list. The whole page is in the history only once, so if
user goes from API root to a resource and clicks all tabs there, a
single click on back button will bring them back to API root.

JIRA: PDC-1011


erichuanggit:
Looks good to me.

PR #118, opened 2015-09-14
Revisit unique constraint for release component

erichuanggit:
Previous release component is uniqued by release, name and global component,
which allows for example to have a release with two release component that
have the same name and different global component.
Change its unique constraint with release and name.

References to release component will be in seperated patch.

JIRA: PDC-1002


ycheng-aa:
Looks good.

simozhan:
Good to me.

PR #119, opened 2015-09-14
Find older compose from other releases

lubomir:
When requesting an older compose with different version of an RPM, the
composes from previous releases of the same product are explored as
well.

When going back, only GA releases are considered. This way, when
starting from compose for r-1.1-updates or r-1.2, composes from r-1.0
and r-1.1 will be considered, but not from r-1.0-updates.

    r-1.0 --------+- r-1.0-updates -----
                  |
                  \- r-1.1 -----+- r-1.1-updates -----
                                |
                                \- r-1.2 -----

JIRA: PDC-1008

### DIFF
--- a/pdc/apps/compose/views.py
+++ b/pdc/apps/compose/views.py
@@ -1165,9 +1165,16 @@ class FindComposeMixin(object):
         current_rpms = set(r.sort_key for r in compose.get_rpms(self.rpm_name))
         # Find older composes for same release (not including this one)
         composes = (Compose.objects
+                    # Get only older composes
                     .exclude(compose_date__gt=compose.compose_date)
-                    .filter(release=compose.release)
+                    # Only composes in the same product
+                    .filter(release__short=compose.release.short)
+                    # Which have the requested rpm in some version
                     .filter(variant__variantarch__composerpm__rpm__name=self.rpm_name)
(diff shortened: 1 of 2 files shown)


pdc/apps/compose/views.py
```diff
@@ -1165,9 +1165,16 @@ def _get_older_compose(self):
...
                     .exclude(compose_date__gt=compose.compose_date)
-                    .filter(release=compose.release)
+                    # Only composes in the same product
+                    .filter(release__short=compose.release.short)
+                    # Which have the requested rpm in some version
                     .filter(variant__variantarch__composerpm__rpm__name=self.rpm_name)
+                    # Keep only composes from the release that requested
+                    # compose belongs to, or GA releases. This way, after R-1.1
+                    # it goes to R-1.0, but not R-1.0-updates.
+                    .filter(Q(release__release_type__short='ga') | Q(release=compose.release))
```
ycheng-aa: What does ga stand for? Thanks.

ycheng-aa:
Looks good.

pdc/apps/compose/views.py
lubomir: General availability.

bliuredhat:
+1

erichuanggit:
Looks good to me.

PR #121, opened 2015-09-15
Fix error message in strict int filter

lubomir:
The filter displayed name of filter as the database filter, not what
user specified.

JIRA: PDC-999


erichuanggit:
Looks good to me.
A unit test would be better.

lubomir:
I added a check to existing test.

erichuanggit:
:+1:

PR #122, opened 2015-09-15
Start work on alternative client

lubomir:
This client will have nicer user interface. The client allows loading
plugins from predefined directories so that implementing internal
extensions is possible.

#### TODO
 * [x] Split help into user help (`--help`) and admin help (`--help-all`).
 * [x] Figure out a way to split the code into multiple modules.
 * [ ] Add unit tests.

### DIFF
--- a/pdc_client/bin/pdc
+++ b/pdc_client/bin/pdc
@@ -20,6 +20,8 @@ PLUGIN_DIRS = [
     os.path.join(os.path.dirname(__file__), '..', 'plugins')
 ]
 
+DATA_PREFIX = 'data__'
+
 
 def get_paged(res, **kwargs):
     """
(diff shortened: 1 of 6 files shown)


pdc_client/bin/pdc
```diff
@@ -0,0 +1,321 @@
...
+import os.path
+import warnings
+import itertools
+import logging
+import imp
+import argcomplete
+
+
+PLUGIN_DIRS = [
+    os.path.join(os.path.dirname(__file__), '..', 'plugins')
```
erichuanggit: what's `plugins` for?

requirements/production.txt
```diff
@@ -12,3 +12,5 @@ django-filter==0.9.2
 django-mptt>=0.7.1
 beanbag>=1.9.2
 django-cors-headers
+argparse
```
erichuanggit: I think both of them need to be added in pdc_client.spec as well.

pdc_client/bin/pdc
lubomir: Supposing the client should be open-source, we need some way to extend it to work with internal parts. This was the easiest way to achieve that. The client finds all python modules in a predefined place, loads the modules and later calls functions in them (`run_hook`). The internal things would just implement the module, drop it into proper place and the client would just work.

requirements/production.txt
lubomir: Correct, but argparse is actually in standard lib, so it should be removed.

erichuanggit:
I like this code structure, loose couple and easy to maintain.

lubomir:
I have rebased the patches on master and restructured the commits so that they are meaningful. It is on pull request #125. So far, I don't have a clear idea on how to reasonably test the code.

PR #124, opened 2015-09-16
Correct warning message

erichuanggit:
Current release component is constraint by release and name but
without global component. It should not contain global component
as unique constraint field.

JIRA: PDC-1002


simozhan:
+1

lubomir:
Is the validator necessary at all? `ModelSerializer` should pick the unique constraint from the model directly.

erichuanggit:
The `UniqueTogetherValidation` class always imposes an implicit constraint that all the fields it applies to are always treated as required.
Since we have `validate_global_component` validate function. Looks like it is not necessary. I will remove it.

lubomir:
:+1:

bliuredhat:
+1

PR #130, opened 2015-09-21
Use helper class in original client

lubomir:
Since the new client has abstraction around BeanBag, this class can be
used in the original client. The PDCClient class is updated in this
patch to support change comment.

There should be no change in the functionality of the old client.
However, without automated tests, it's hard to say for sure.


erichuanggit:
Looks good to me.

PR #131, opened 2015-09-22
Add tests to client

lubomir:
The main part is the first commit, which adds base classes for the tests.

Generally, the tests are able to verify the client connects to end-points it is supposed to connect to, does not connect to anything else and that it has the correct output.

Since the client code is now included in the coverage measure, it is expected to drop a bit.


erichuanggit:
Looks good to me.
Well done.

PR #134, opened 2015-09-24
PDC-1020 Create github pages with rendered docs

lao605:
now the gh-pages branch is available. It uses the very project mentioned in the desc ：sphinx-deployment

it requires 2 steps to update the docs:

1. run `make html` in `docs` directory
2. run `make deploy_gh_pages` in `docs` directory

when pushed, travis-CI will be triggered to auto-regenerate sphinx docs and push the generated html to gh-pages branch.

and the result can be viewed in http://release-engineering.github.io/product-definition-center/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ addons:
 # any part of kobo that would actively use koji, se we just delete the
 # offending import.
 install:
+    - git config --global user.name  "Travis CI"
+    - git config --global user.email "pdc@product-definition-center.com"
+    - export REPO_URL_GITHUB="https://$GH_TOKEN@github.com/$GH_REPO.git"
     - pip install -U pip wheel coveralls
     - pip install -r requirements/devel.txt
     - sed -i -E '/import (koji|rpm)/d' $(python -vc 'import kobo.rpmlib' 2>&1 | grep 'import kobo.rpmlib' | awk '{print$NF}' | sed 's/pyc$/py/')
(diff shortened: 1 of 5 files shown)


lubomir:
I don't see any reason why the `docs/build` directory should be in version control, that seems like a mistake. Also the `.deploy_heroku` stuff is quite useless to use, so let's get rid of it. Removing the readme and license (just the file, not the copyright notice in actual files) might also be a good idea.

One modification I would love to see in the Makefile is to make sure the `date` command is always invoked with `LANG=C date` so that no matter what the current locale is, the text is in English.

Please squash the patches.

The travis configuration is actually not completely correct. It should only deploy the docs if it was a push to master. This way it would re-deploy on every proposed pull request.

lao605:
@lubomir 

Do we need to discuss the email address next week or just take that `travis@testing-worker-linux-docker-b6df935f-3197-linux-3.prod.travis-ci.org`?

http://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables shows that all the VM images in Travis CI share some common environment including `LANG=en_US.UTF-8`. What do you mean by `current locale`? Do you attempt to run Travis CI locally?

lubomir:
I guess the e-mail address does not really, it's just that git complains if it is not set. But that is a minor thing.

When running on travis, the locale is set correctly. Originally I mean to fix the case when someone deploys the docs manually from their local machine. For example if I went to the docs directory and executed `make deploy`, it would log date as *Čt zář 24 09:35:50 UTC 2015* which is probably not desirable.

lao605:
yes I am just looking for that script! thanks~

lubomir:
:+1:

PR #135, opened 2015-09-24
Add build images plugin.

ycheng-aa:
JIRA: 1025


ycheng-aa:
I can't run client test successfully even in master branch. I will add unit tests after I fixed it.

erichuanggit:
Looks good to me.
BTW, current client tests work for me.

lubomir:
If you need help with the tests, ping me.

I'm not a big fan of the way rpms and archives are specified. The expected format does not show up in help. I don't know how it could be made more user friendly and still allow specifying multiple objects. Anyway, I would prefer the input to be JSON instead of Python literal.

ycheng-aa:
@lubomir  Thanks. I haven't figured out a simple way for rpms and archives to present multiple objects with multiple fields. All I thought was too complicated to use in real. I will change it to JSON first and think it a little more.

lubomir:
@ycheng-aa The only idea I have is to have a separate command `build-image-add-X`, but that would need adding the items one-by-one. Anyway, maybe we don't really need this, because the images should be imported in some other way? Just list and info would be enough then. @sochotnicky, any preference here?

ycheng-aa:
@erichuanggit helped me compare the env to investigate the unit test failed problem. It's because I use Fedora 22 with mock version 1.3.0. After I changed it to 1.0.1, it worked. The error was about "assert_called". Not sure if it's a potential issue.

lubomir:
@ycheng-aa Well, we'll have to fix some day, either restrict the version or make it work with both versions.

bliuredhat:
+1

ycheng-aa:
Change to JSON instead of Python literal. Before  @sochotnicky make the decision,  I keep the POST and PATCH here and added the tests.

lubomir:
Looks good. The error handling of retrieving single build image could be removed by accessing the detail URL directly, but that part of the client will need to be improved anyway, so it is no big deal.

erichuanggit:
Looks good to me.

sochotnicky:
@ycheng-aa Generally for content of images I don't think we need to have an very user-friendly way to do it. Expectation is that the POST API is mostly used by other tools/scripts so not a big deal. Current version looks good otherwise

lubomir:
Cool, let's merge it then. However, the question for future is: does it make sense to add stuff to *user*-friendly client if users are not expected to use it?

sochotnicky:
@lubomir Good point - how about we make only the filtering part in the client and leave the editing to lower-level

ycheng-aa:
Removed patch and create function and corresponding test. Just keep the build-image-list and build-image-info.

lubomir:
:+1:

ycheng-aa:
Merged.

PR #136, opened 2015-09-25
Fix `extract_arguments` in plugin_helpers.

xychu:
We need to rebuild dict for group args, like:
from
```
    group__keyA
    group__keyB
```
to
```
    {'group': {
        'keyA': value,
        'keyB': value}
    }
```
or even with subgroup
```
    group__subgroup-a__key
    group__subgroup-b__key
```
to
```
    {'group': {
        'subgroup-a': {'key': value},
        'subgroup-b': {'key': value}}
    }
```

### DIFF
--- a/pdc_client/plugin_helpers.py
+++ b/pdc_client/plugin_helpers.py
@@ -130,9 +130,18 @@ def extract_arguments(args, prefix=DATA_PREFIX):
     for key, value in args.__dict__.iteritems():
         if key.startswith(prefix) and value is not None:
             parts = key[len(prefix):].split('__')
-            d = data
-            for p in parts[:-1]:
-                d[p] = {}
-                d = d[p]
-            d[parts[-1]] = value if value != '' else None
+            if len(parts) == 1:
+                data[parts[0]] = value if value != '' else None
+            elif len(parts) > 1:
+                temp_dict = dict()
+                for p in reversed(parts):
+                    if not temp_dict:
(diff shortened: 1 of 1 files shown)


lubomir:
This was actually already supposed to work, I just forgot to test with multiple keys in the same group. The alternative fix is this:
```
--- a/pdc_client/plugin_helpers.py
+++ b/pdc_client/plugin_helpers.py
@@ -132,7 +132,6 @@ def extract_arguments(args, prefix=DATA_PREFIX):
             parts = key[len(prefix):].split('__')
             d = data
             for p in parts[:-1]:
-                d[p] = {}
-                d = d[p]
+                d = d.setdefault(p, {})
             d[parts[-1]] = value if value != '' else None
     return data
```

xychu:
If I understand it correctly, we need to nested the result dict based the group information,
so for rpm dependencies it will be
```
    {'dependencies': {
        'requires': [...],
        'provides': [...],
        ...}
    }
```

lubomir:
That is exactly how it should work. I added some comments explaining the code (on a branch): https://github.com/release-engineering/product-definition-center/blob/allow-nested-args/pdc_client/plugin_helpers.py#L126-L149

xychu:
Then we should merge your branch :-)
Don't know why I did not get the expected result before.

lubomir:
@xychu Ok, merged (now with a test case).

PR #137, opened 2015-09-25
Add component client plugins

xychu:
Including Global Component and Release Component.


lubomir:
Generally looks good to me, but CLI arguments with underscores are not consistent with other plugins and generally I think dashes are more common (i.e. `--include-inactive-release` instead of `--include_inactive_release`).

bliuredhat:
+1

xychu:
Updated the `--include-inactive-release`.
(Force updated, my 'bad' habit to amend commit instead of adding new ones to make it easier to merge but harder to review. Hope this is the last time.)

lubomir:
I still think the arguments should not be using underscores. There are some left. When adding it to parser, setting up `dest` should not be necessary (`plugins/component.py:144`), the resulting value should be available with underscores in the name automatically.

erichuanggit:
Looks good to me.

lubomir:
:+1:

PR #139, opened 2015-09-29
Add tracking OSBS metadata

lubomir:
Currently `autorebuild` only. Please see commit message on the second commit for more details.

### DIFF
--- a/pdc/apps/component/migrations/0004_auto_20150708_0746.py
+++ b/pdc/apps/component/migrations/0004_auto_20150708_0746.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 
 from django.db import models, migrations
 import django.db.models.deletion
-from pdc.apps.component.models import ReleaseComponentType
 
 
 def create_release_component_types(apps, schema_editor):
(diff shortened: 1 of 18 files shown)


pdc/apps/component/migrations/0004_auto_20150708_0746.py
```diff
@@ -14,6 +13,7 @@ def create_release_component_types(apps, schema_editor):
         'iso',
         'composite'
     ]
+    ReleaseComponentType = apps.get_model('component', 'ReleaseComponentType')
```
erichuanggit: what's the benefit but not importing ReleaseComponentType directly? 
I think this migration file won't be applied in latest database.

erichuanggit:
Generally looks good to me.

pdc/apps/component/migrations/0004_auto_20150708_0746.py
lubomir: On latest database, the migration is probably already applied, so it should not matter. Only with new database this migration can't be applied.

It is not possible to import directly, as that would import the latest model (in this case the one with additional `has_osbs` field). We need the historical version here.

PR #140, opened 2015-09-30
Make login URL customizable

lubomir:
Load LOGIN_URL from settings and use that. Kerberos auth is no longer
hardcoded, but it is still the default settings value.


ycheng-aa:
Looks good.

erichuanggit:
Looks good to me.

pypingou:
Tested at https://pdc.fedorainfracloud.org/ seems to work fine :)

PR #141, opened 2015-09-30
Add SAML2 support with the help of mod_auth_mellon

puiterwijk:
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>


pdc/apps/auth/middleware.py
```diff
@@ -48,9 +49,12 @@ def process_request(self, request):
...
-        user = auth.authenticate(remote_user=username)
+        user = auth.authenticate(remote_user=username, request=request)
         if user:
             # User is valid.  Set request.user and persist user in the session
             # by logging the user in.
             request.user = user
-            auth.login(request, user)
+            request.session['auth_backend'] = user.backend
+            backend = load_backend(user.backend)
+            if not hasattr(backend, 'save_login') or not backend.save_login:
```
puiterwijk: This line is here because it will allow mod_auth_mellon to control the login state.
The advantage of this is that is that we get Single Logout for free.

puiterwijk:
Hmm, note that this will require either a single load balancer with mod_auth_mellon or a single application server. This does NOT scale to multiple app servers, since mod_auth_mellon doesn't share its cache between multiple servers.

pdc/apps/auth/middleware.py
lubomir: I probably don't understand that. `KerberosUserBackend` does not have the `save_login` attribute, so condition is true. `AuthMellonUserBackend` has `save_login = False`, so with the negation the condition is also true. Is this intended?

Essentially, the condition is equivalent to `not getattr(backend, 'save_login', False)`, while I think it would make more sense as `getattr(backend, 'save_login', True)`. What am I missing?

lubomir:
I don't think restricting to single app server is a big deal right now. Ideally, this should be documented (comment on the backend would be enough).

Regarding the name parsing: the only reason we have split names is because that is how Django does it. PDC only ever uses full name. Since we override the user class anyway, we could use a single string field for user names in the future and simplify this part.

pypingou:
> This does NOT scale to multiple app servers, since mod_auth_mellon doesn't share its cache between multiple servers.

Do we know where the cache is stored? If so maybe we could make it work with say a gluster volume?

puiterwijk:
Yes, the cache is only stored in server memory. We are now working upstream with mod_auth_mellon to add a distributed cache mechanism to fix this case, but for the time being the entire session cache is stored in server memory.

pdc/apps/auth/middleware.py
puiterwijk: I am trying to default to True for anything except AuthMellonBackend.
Every backend still needs to save logins, only the mod_auth_mellon backend doesn't, because mod_auth_mellon will handle sessions.

puiterwijk: Agreed though, not getattr(backend, 'save_login', False)  is easier to read. Will do that instead.

puiterwijk: You were entirely correct, and I had that not the wrong way around.
I now have getattr(..., True).

lubomir:
Looks good to me. Merged as afad4897a57c46c36271a63d10305e76469c1515.

PR #155, opened 2015-10-13
Add filter dist_git_branch for release component

erichuanggit:
JIRA: PDC-1038

### DIFF
--- a/pdc/apps/component/filters.py
+++ b/pdc/apps/component/filters.py
@@ -155,12 +155,9 @@ class ReleaseComponentFilter(ComposeFilterSet):
 
     @value_is_not_empty
     def filter_by_dist_git_branch(self, qs, value):
-        result = qs.filter(dist_git_branch__in=value)
-        if result:
-            return result
-        else:
-            q = Q(release__releasedistgitmapping__dist_git_branch__in=value) and Q(dist_git_branch__isnull=True)
-            return qs.filter(q)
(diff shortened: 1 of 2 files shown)


lubomir:
Mostly looks fine. I think it would be better if the test created the mapping directly, since it is not supposed to create the update API. But that is a minor concern. It can be merged as is.

erichuanggit:
Just realized that dist_git_branch could also be inherited from release. Need to revise the change.

lubomir:
Great point about inheritance. The update is not completely correct, though. Example: release has dist git branch *foo*, its component has dist git branch *bar*. Filter components with branch *foo*, the output contains the component which overrides it.

erichuanggit:
Right. will fix in coming patch.

lubomir:
The fix is still not correct. Admittedly, this is not a common use case, but since we don't prohibit such data, it should be handled correctly.
 * release has branch foo
 * component1 has no branch, it inherits foo from release
 * component2 has branch set to foo explicitly

Filtering components by branch foo shows only component2.

I think the previous version would work with filtering by the following expression: get all components with given branch, or components with unset dist git branch belonging to release with that branch.
```
Q(dist_git_branch__in=value) | Q(release__releasedistgitmapping__dist_git_branch__in=value,
                                 dist_git_branch__isnull=True)
```

erichuanggit:
I will add more test cases. Thanks

ycheng-aa:
:+1:

lubomir:
I can't think of any more edge cases, so :+1: from me.

PR #156, opened 2015-10-13
Add support for storing partner information

lubomir:
When delivering data to customers, some partners may have access to
various data before general availability. It is necessary to store this
kind of information. This patch adds storage for partner metadata and
allows linking partners to specific Variant.Arches on releases.

Partner data can be completely edited through the API as well as the
mapping. Partner types can only be modified in Django admin.

JIRA: PDC-1048

### DIFF
--- a/pdc/apps/partners/serializers.py
+++ b/pdc/apps/partners/serializers.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2015 Red Hat
+# Licensed under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+#
+from rest_framework import serializers
+
+from pdc.apps.common.serializers import StrictSerializerMixin
+from pdc.apps.common.fields import ChoiceSlugField
+from pdc.apps.release import models as release_models
+from . import models
+
+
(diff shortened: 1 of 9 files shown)


pdc/apps/partners/serializers.py
```diff
@@ -0,0 +1,62 @@
...
+                variant__variant_uid=variant_uid,
+                arch__name=arch
+            )
+        except release_models.VariantArch.DoesNotExist:
+            raise serializers.ValidationError(
+                'No VariantArch for release_id=%s, variant_uid=%s, arch=%s'
+                % (release_id, variant_uid, arch)
+            )
+        try:
+            models.PartnerMapping.objects.get(**attrs)
```
erichuanggit: why not using `self.instance` to check if partners mapping already exists? since both `create` and `update` validation will run into this function.

lubomir:
I tried to simplify the serializer. We don't have to care update for the mapping, the only meaningful actions are create and delete.

By the way, it would probably make sense to move `VariantArch` field to `release` app and use it for `RepoSerializer` as well, but that would be a breaking change. I will file an issue about it.

erichuanggit:
Looks good to me.

xychu:
:+1:

PR #157, opened 2015-10-13
Give better error message when importing incorrect composeinfo

lubomir:
It can easily happen that the release version is bumped but not its
integrated layered products. The import fails in such case. Now there is
a clear error message explaining what the problem is.

JIRA: PDC-1033


erichuanggit:
Looks good to me.

PR #159, opened 2015-10-13
Add documentation to repo retrieve

lubomir:
This action was added when repos got identifiers. It was not
documented, though.

PDC-1071


erichuanggit:
Looks good to me.

PR #160, opened 2015-10-13
Mark required arguments in client create

lubomir:
The usage should not wrap required arguments in brackets. The fixed
create commands are for release, rpm, global and release components.

The test for creating RPMs was updated to list all required arguments.

JIRA: PDC-1084, PDC-1079


erichuanggit:
Looks good to me.

PR #161, opened 2015-10-14
Add link to change list to user menu

bliuredhat:
When user is logged in, the changes are accessible through this link.

JIRA: PDC-1052

### DIFF
--- a/pdc/templates/base_generic.html
+++ b/pdc/templates/base_generic.html
@@ -67,8 +67,11 @@
                 <li>
                   <a href='{% url 'auth/logout' %}?next={{ request.path }}'>{% trans 'Log-out' %}</a>
                 </li>
-                {% if user.is_superuser %}
                 <li class="divider"></li>
+                <li>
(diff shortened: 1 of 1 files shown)


lubomir:
I would prefer the link after the delimiter, so that when user is logged in, the order would be Profile, Log out, Delimiter, Changes (followed by Admin if superuser).

The commit message should be reformatted, so that there is first line at most 50 characters long, followed by blank line and more detailed description. Current message breaks `git log --oneline` a bit. I would suggest something like

```
Add link to change list to user menu

When user is logged in, the changes are accessible through this link.

JIRA: PDC-1052
```

bliuredhat:
@lubomir ,  good ideal and I rewrite it ~!
Thank you~!

xychu:
LGTM

lubomir:
In my opinion, the changes link should be visible for all authenticated users, not just admins.

lubomir:
Perfect. Please rebase on master and merge.

PR #163, opened 2015-10-14
Allow resetting product version

lubomir:
Resetting base products was implemented in previous patch. Product
version should have been updated to work the same way, but was
forgotten. Test is added.

JIRA: PDC-1095


erichuanggit:
Looks good to me.

PR #164, opened 2015-10-14
Rename release type field in client output

lubomir:
All other places refer to the value as "release type". It is better to
keep the work release in the label in the client.

JIRA: PDC-1068


erichuanggit:
Looks good to me.

simozhan:
looks good to me.

PR #165, opened 2015-10-14
Fix dependency documentation

lubomir:
The input should be a list of strings, not a single string.

JIRA: PDC-1029


bliuredhat:
looks good~, +1

PR #167, opened 2015-10-15
Update docs in GlobalComponent views.

xychu:
Make /update fields consistent with /create and new client arguments.


lubomir:
Looks good to me.

erichuanggit:
Looks good to me.

PR #169, opened 2015-10-15
Allow empty directories for partners

lubomir:
The original issue was not clear on ftp_dir and rsync_dir fields. The
were always required. Turns out the fields should not be. This patch
changes them to accept empty string as possible value.

JIRA: PDC-1048


erichuanggit:
Looks good to me.

PR #170, opened 2015-10-15
Allow having multiple contacts with same role for one component

ycheng-aa:
Add a field count_limit to ContactRole to define the limit of
this role for each component.
The value default is 1, and define 0 as unlimited. When exceed
the limit, a ValidationError exception will be raised.

JIRA: PDC-1056

### DIFF
--- a/pdc/apps/contact/models.py
+++ b/pdc/apps/contact/models.py
@@ -173,7 +173,7 @@ class GlobalComponentContact(ValidateRoleCountMixin, models.Model):
                                   on_delete=models.PROTECT)
 
     def __unicode__(self):
-        return u'%s: %s: %s' % (unicode(self.component), self.contact_role, unicode(self.contact))
+        return u'%s: %s: %s' % (unicode(self.component), self.role, unicode(self.contact))
 
     class Meta:
         unique_together = (('role', 'component', 'contact'), )
--- a/pdc/apps/contact/models.py
+++ b/pdc/apps/contact/models.py
@@ -13,6 +13,8 @@ from django.forms.models import model_to_dict
 class ContactRole(models.Model):
 
     name = models.CharField(max_length=128, unique=True)
+    count_limit = models.IntegerField(default=1)
+    UNLIMITED = 0
 
     def __unicode__(self):
         return u"%s" % self.name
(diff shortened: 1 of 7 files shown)


lubomir:
Minor nitpick: the `contact-roles` API should should probably show the limit.

A bigger concern is that now the constraint is handled in the serializer. In the future, if we are creating contacts anywhere else in the code, the role count would not be checked. For this reason, I would prefer the check in `clean` method on the actual model. The check would be executed automatically by `pre_save` signal. It would even simplify the implementation as it would not need to handle create and update differently. In order not to have to duplicate code, a single mixin could implement the `clean` method and be used for both models.

lubomir:
Looks pretty good to me now. Please add a period at the end of error message for consistency.

The code could simplified a bit in my opinion:
 * get all matching contacts
 * exclude current one if it has `pk`
 * perform count check

This way it would nicely fit into single method.

I also think the check for unlimited should be done before querying the database. It would save us one query in some cases.

pdc/apps/contact/models.py
```diff
@@ -171,10 +186,10 @@ class ReleaseComponentContact(models.Model):
                                   on_delete=models.PROTECT)
 
     def __unicode__(self):
-        return u'%s: %s: %s' % (unicode(self.component), self.contact_role, unicode(self.contact))
+        return u'%s: %s: %s' % (unicode(self.component), self.role, unicode(self.contact))
```
erichuanggit: should add `unicode` for `role` as well?

pdc/apps/contact/models.py
```diff
@@ -8,11 +8,14 @@
 from django.db import models
 from django.db.models.query import QuerySet
 from django.forms.models import model_to_dict
+from django.core.exceptions import ValidationError
 
 
 class ContactRole(models.Model):
 
     name = models.CharField(max_length=128, unique=True)
+    count_limit = models.IntegerField(default=1)
```
erichuanggit: how about use `count_contact_limit`? for my first glimpse, I understood it as the count limit of Role :-)

erichuanggit:
Generally looks good to me.

pdc/apps/contact/models.py
lubomir: Adding a comment seems better that changing the name. Maybe [help text](https://docs.djangoproject.com/en/dev/ref/models/fields/#help-text) or just mention in docstring would be enough?

pdc/apps/contact/models.py
ycheng-aa: Fixed in0982fbd94

pdc/apps/contact/models.py
ycheng-aa: Added help text in 0982fbd94.

lubomir:
Looks good.

ycheng-aa:
Merged.

PR #172, opened 2015-10-16
Fix product_id filter on content delivery repos

lubomir:
The value should be an integer.

JIRA: PDC-1104


erichuanggit:
Looks good to me.

PR #173, opened 2015-10-19
Wrong input filed name will not be updated.

simozhan:
JIRA: PDC-1099


ycheng-aa:
Looks good.

erichuanggit:
Prefer to @lubomir's by extending StrictSerializerMixin, since we already have such common handler.

lubomir:
:+1:

PR #174, opened 2015-10-19
Sort arches in web view

lubomir:
Up to now, the list was sorted based on the order in which arches were
created. With manually added ones, the order was not correct.

JIRA: PDC-1107


erichuanggit:
Looks good to me.

PR #175, opened 2015-10-19
Improve help output for create commands

lubomir:
Required arguments for create commands are now displayed in a separate
block. For update action, all the arguments end up in one block.

JIRA: PDC-1079


erichuanggit:
Looks good to me.

PR #176, opened 2015-10-19
Add example script to find components with duplicate contacts

lubomir:
While this script is not terribly useful for day-to-day life, it can
serve as an example on how to use the client.

JIRA: PDC-1105

### DIFF
--- a/examples/find-duplicated-contacts.py
+++ b/examples/find-duplicated-contacts.py
@@ -1,0 +1,35 @@
+#! /usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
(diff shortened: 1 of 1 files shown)


ycheng-aa:
Should the "duplicit" in the title be "duplicate"?

lubomir:
@ycheng-aa Fixed, thanks.

erichuanggit:
@lubomir I remember you use `duplicit` in multiple place in the unit test. Not sure if it is one word from your native language :-)
:+1:

lubomir:
@erichuanggit Yeah, that's it. Thanks for pointing this out, apparently that's a mistake I make often.

PR #177, opened 2015-10-20
Serializer documentation fix + tests

lubomir:
First patch fixes a bug about documentation, second patch adds some tests for serializer documenatation generator. The patches are not really closely related, but the second depends on the first one (due to changed output).


erichuanggit:
Looks good to me.

PR #178, opened 2015-10-21
Make rpms document consistant with response

erichuanggit:
JIRA: PDC-1030


lubomir:
:+1: This is not just correct, it is definitely more helpful that it was before.

ycheng-aa:
Looks good.

simozhan:
+1

PR #179, opened 2015-10-21
Automatically regenerate schema diagrams

lubomir:
On each push to master, before documentation is deployed, the diagrams
are regenerated.

There is actually an issue with the generated documentation: Travis only
checks out latest 50 commits. If the latest python-pdc-* tag is older
than that, getting version will fail.

JIRA: PDC-1110


erichuanggit:
Looks good to me.
If we tag every SP, I think it won't be problem.

lubomir:
That could work. However, we need to figure out some consistent way of using the tags. Currently, the version is base only on tags created by tito (and it's actually not correct – dev server shows version as  0.1.0-1.79.g2c7e619). I think we should switch it back to using the manual tags `v0.*.*` and just make sure that tito builds are done from that commit.

ycheng-aa:
:+1:

bliuredhat:
+1

PR #180, opened 2015-10-21
Fix contact role count limit bug.

ycheng-aa:
Decrease count limit for a contact role to a number
lower than existing maximum role count is not
allowed any more.


simozhan:
looks good to me.

lubomir:
:+1:

bliuredhat:
+1

PR #181, opened 2015-10-23
 Update the release component serializer file

bliuredhat:
The code in component serializer  is not reachable,
and remove the unreachable code.

JIRA: PDC-1053

### DIFF
--- a/pdc/apps/component/serializers.py
+++ b/pdc/apps/component/serializers.py
@@ -7,7 +7,6 @@ import json
 
 from django.contrib.contenttypes.models import ContentType
 from django.core.urlresolvers import reverse
-from django.shortcuts import get_object_or_404
 from django.utils import six
 from django.utils.text import capfirst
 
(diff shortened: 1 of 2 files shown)


pdc/apps/component/serializers.py
```diff
@@ -373,32 +373,6 @@ def to_internal_value(self, data):
...
-        if not isinstance(value, Release):
-            if isinstance(value, dict):
-                release_id = value['release_id']
-            else:
-                release_id = value
-            if release_id is None or release_id.strip() == "":
-                self._errors = {'release': 'This field is required.'}
-                return
-            release = get_object_or_404(Release, release_id=release_id)
-            if not release.is_active():
```
xychu: If we need to keep the check of "Not create release component with an inactive release", we'd better keep `validate_release` here.
The reason that test [here](https://github.com/release-engineering/product-definition-center/blob/master/pdc/apps/component/tests.py#L1747-L1755) get passed is because the 'releases' in data is should be 'release'.
Please correct other tests of `ReleaseComponent` that with 'releases' in data as well.

lubomir: @xychu That's a good point. However, the validation still needs to be updated, because the `value` argument is always an instance of `Release`. Raising `ValidationError` would also be a preferable way of failing.

Please note that for bulk update, the key is actually called `releases`, so this is quite confusing.

xychu: Agreed, the `isinstance

bliuredhat:
@lubomir  and @xychu ; thank you so much and I will rewrite the script as followed your suggestion~!

bliuredhat:
@lubomir and @xychu ,  whether this branch can be merged to master branch ~?

lubomir:
This looks good to me.

PR #185, opened 2015-10-23
Return JSON 404 response from Django

lubomir:
When Django generates 404 response, it would always return the rendered
template, which is not nice as the client might be expecting JSON. This
patch adds a check for the Accept HTTP header. If it mentions
application/json MIME type, JSON object is returned instead.

This is automatically picked by pdc_client as it already adds the
header.

JIRA: PDC-1106

### DIFF
--- a/pdc/apps/common/handlers.py
+++ b/pdc/apps/common/handlers.py
@@ -13,6 +13,9 @@ import sys
 import logging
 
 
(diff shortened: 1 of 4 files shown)


erichuanggit:
Looks good to me.

bliuredhat:
+1

ycheng-aa:
:+1:

PR #186, opened 2015-10-26
Improve performance of the web UI

lubomir:
Many of the pages performed lots of database queries. By selecting and
prefetching related data, that number can be brought down significantly.

This patch required rewriting the logic for determining if product or
product version is active. Now it  able to take advantage of the
preloaded data. A test case was added to make sure it works correctly.

JIRA: PDC-1132

### DIFF
--- a/pdc/apps/release/models.py
+++ b/pdc/apps/release/models.py
@@ -72,24 +72,23 @@ class Product(models.Model):
 
     @property
     def active(self):
-        releases = Release.objects.filter(active=True).filter(product_version__product=self)
-        return bool(releases)
+        return self.active_release_count > 0
 
     @property
     def product_version_count(self):
-        return ProductVersion.objects.filter(product=self).count()
+        return self.productversion_set.count()
 
     @property
     def active_product_version_count(self):
-        return ProductVersion.objects.filter(product=self).filter(release__active=True).distinct().count()
+        return sum(1 for pv in self.productversion_set.all() if pv.active)
 
     @property
     def release_count(self):
-        return Release.objects.filter(product_version__product=self).count()
+        return sum(pv.release_count for pv in self.productversion_set.all())
 
     @property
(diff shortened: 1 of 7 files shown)


erichuanggit:
Looks good to me.

pdc/apps/release/models.py
```diff
@@ -236,6 +234,19 @@ def version_sort_key(self):
...
 
+    @property
+    def integrated_release_variants(self):
+        """
+        Returns mapping from Variant-UID to release instance from which the
+        variant is integrated.
+        """
+        if not hasattr(self, '_integrated_release_variants'):
+            self._integrated_release_variants = {}
+            for release in self.integrated_releases.all():
```
ycheng-aa: Should here also use prefetch_related to reduce DB query number, something like:
for release in self.integrated_releases.prefetch_related('variant_set')?

pdc/apps/release/models.py
```diff
@@ -72,24 +72,23 @@ def __unicode__(self):
...
 
     @property
     def release_count(self):
-        return Release.objects.filter(product_version__product=self).count()
+        return sum(pv.release_count for pv in self.productversion_set.all())
 
     @property
     def active_release_count(self):
-        return Release.objects.filter(product_version__product=self).filter(active=True).distinct().count()
+        return sum(pv.active_release_count for pv in self.productversion_set.all())
```
ycheng-aa: Should we prefetch_related('release_set') for productversion_set to reduce DB query number?

ycheng-aa:
Looks good.  Add comments for some places that I am not sure if it could improve a little.

simozhan:
Looks pretty good to me. 
thumbup

lubomir:
@ycheng-aa Both cases you mentioned are actually covered in corresponding view. They are not used anywhere else.

PR #187, opened 2015-10-26
Role count limit improvements

lubomir:
Accept `"unlimited"` as valid input for role count limit and fix the examples.


ycheng-aa:
Looks good.

Maybe we should also change count_limit definition in model to PositiveIntegerField to not allow negative number.

erichuanggit:
Looks good to me. 
If update `count_limit` data type, I would prefer to `PositiveSmallIntegerField`, the count of contact for specified contact_role should be less than 32767.

lubomir:
Very good point, guys. I will change the field type.

simozhan:
looks good

PR #189, opened 2015-10-28
Rename resource names

erichuanggit:
To be consistent with other end-points, rename below three ones:
* content-delivery-content-category -> content-delivery-content-categories
* content-delivery-content-format -> content-delivery-content-formats
* content-delivery-service -> content-delivery-services
The old names remain available for backwards compatibility, but has been
marked deprecated.

JIRA: PDC-1027


simozhan:
looks good to me

lubomir:
I think it would be nice to change the router `base_name`s for the deprecated end-points. With the patch there are duplicate `base_name`s in there, and it may not cause problems, but it is confusing for readers of the code.

erichuanggit:
@lubomir Very nice proposal. Since the deprecated end-point will be remove in future. I did not think about more.

lubomir:
Looks fine.

erichuanggit:
@lubomir Thanks. `deprecated` should be the right one, will fix.

ycheng-aa:
Looks good.

PR #192, opened 2015-10-30
Allow rpm dependencies to be duplicated.

simozhan:
Simply deleted the unique constraint. If the dependency data is very big, maybe add a relationship table between rpm and dependency is better?

### DIFF
--- a/pdc/apps/package/migrations/0006_auto_20151030_0845.py
+++ b/pdc/apps/package/migrations/0006_auto_20151030_0845.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('package', '0005_auto_20150907_0905'),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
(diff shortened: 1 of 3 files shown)


erichuanggit:
I think what @lubomir proposes is ignoring duplicated dependency and storing only once. But current change will store multiple.

ycheng-aa:
From JIRA, "The goal is that anyone can take rpm dependencies and shove them into PDC without having to go through them and doing uniq(sorted()) on them.". It looks like ignoring duplicate dependencies could work. We unique it first and store. If we want record the dependencies precisely, maybe just add a field to record each dependency repetition number? I am not sure. At least I think we don't have to restore the whole information repeatedly.

simozhan:
When we import the rpm, some information is missing. For example in gok.spec
Requires(pre): GConf2
Requires(post): GConf2
I am just thinking if it makes sense to allow duplicates for easy extending Dependency in the future?

lubomir:
In an ideal world I would prefer to remove the duplicates. However, I think the proposed patch with removed unique constraint is fine. It should not cause much problems with amounts of data, as there will not be that many duplicates, and will save space by not using an index for the constraint.

I don't know if we need to handle different kinds of requires (pre versus post). If we do, that would be a different task, I think.

PR #194, opened 2015-11-03
 Restucture the webUI

bliuredhat:
Move the base-product link from 'Releases'  to  'Products' in WebUI menu,
 At the same time, update the release view.py file;
 and remove the unreachable file.

 JIRA: PDC-1091


xychu:
LGTM, and I believe you can do better on the commit msg. :-)
Try something like "Move 'base product' from 'Releases' to 'Product' in WebUI menu.", keep it short is good but more specific is always encouraged.

bliuredhat:
@xychu,  Thank you~!

lubomir:
:+1:

PR #195, opened 2015-11-03
Add order to queryset to ensure correctness of pagination result.

lao605:
Originally, rows of querysets is in unspecified order if sorting is not chosen.
This causes problem when django pagination is used.
It can only be guaranteed if the sort type if explicitly chosen.

JIRA: PDC-1063


erichuanggit:
Looks good to me.

PR #196, opened 2015-11-04
Reduce number of database queries

lubomir:
These are minor improvements that improved performance while working on the partners ui. I'm not completely sure the approach from the first commit is the best option. Adding `prefetch_related` unconditionally broke the tests.

### DIFF
--- a/pdc/apps/release/views.py
+++ b/pdc/apps/release/views.py
@@ -758,7 +758,7 @@ class ReleaseVariantViewSet(ChangeSetModelMixin,
         # Preloading with a separate request could be actually be slow if there
         # is no filter, so it's done only on list request with paging enabled.
         queryset = self.queryset.select_related('variant_type', 'release')
-        if self.action == 'list' and self.paginator.get_page_size(self.request):
+        if self.is_filtered_list():
             queryset = queryset.prefetch_related('variantarch_set__arch')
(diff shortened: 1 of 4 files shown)


pdc/apps/release/views.py
```diff
@@ -748,11 +748,20 @@ class ReleaseVariantViewSet(ChangeSetModelMixin,
...
     deleting a single variant as well as in bulk operations.
     """
-    queryset = models.Variant.objects.all().order_by('id')
+    queryset = models.Variant.objects.order_by('id')
     serializer_class = ReleaseVariantSerializer
     filter_class = filters.ReleaseVariantFilter
     lookup_fields = (('release__release_id', r'[^/]+'), ('variant_uid', r'[^/]+'))
 
+    def get_queryset(self):
+        # Selecting related fields (using a JOIN statement can be done always).
```
erichuanggit: I think this case not only in release variant, why not move to `common`?

lubomir:
I have added a commit which moves part of the code to `StrictQueryParamMixin`. I'm not sure it is a good change, though.

ycheng-aa:
@lubomir Should we continue with pull request?

lubomir:
Let's abandon this. It is not really a problem right now.

PR #197, opened 2015-11-04
Remove Arches and Sigkeys items from menu

bliuredhat:
This task is to remove "Arches" and "Sigkeys" items from
main menu.

At the same time, remove arch_list and sigkey_list files
, and update the view&&urls files.

JIRA: PDC-1091


lubomir:
Yup, I think this removes everything it should. :+1:

erichuanggit:
Looks good to me.

simozhan:
looks good to me

PR #198, opened 2015-11-04
Modify default CORS settings

lubomir:
Sending credentials is now enabled by default. The change comment header
is whitelisted and can be used from UIs.

A bunch of explanatory comments was added to the settings file with
pointers on what might be reasonable to override in local settings.

JIRA: PDC-1088


pdc/settings.py
```diff
@@ -203,10 +203,38 @@
 LDAP_GROUPS_DN = "ou=groups,dc=example,dc=com"
 LDAP_CACHE_HOURS = 24
 
+
+#
 # CORS settings
+#
+
+# The requests can come from any origin (hostname). It this is undesirable, use
```
erichuanggit: It -> If

erichuanggit:
Generally looks good to me.

PR #199, opened 2015-11-04
Use popovers instead of dialogs for help

lubomir:
The popovers are triggered by hovering over the icon. Once the cursor
leaves the icon, the popover is hidden. The only exception is if the
cursor moved over the actual popover, in which case it will disappear
only after the cursor left it.

JIRA: PDC-1115


erichuanggit:
Except for that the font size is too big in the tooltip, other looks good to me.

lubomir:
@erichuanggit It uses same font size as the rest of the web ui. Maybe UX team can help with the design.

erichuanggit:
As talked to UX team, they also suggest making the font smaller in the tooltip. Below is what we tried together in my local: 
```
.popover-content{
    font-size:12px;
}
```

xychu:
Looks good to me.

PR #201, opened 2015-11-05
Split routers.py file

lao605:
Routers are now configured in apps.
Because of dependency problem, apps normally can't be disabled as easily as expected except standalone ones.


lubomir:
Looks good to me.

It's true that some applications can not be disabled, because others depend on them. But for example `partners`, `osbs` or `usage` can easily be removed (well, for `usage` the middleware must be disabled as well). Once PDC-1050 is done, it should also be possible to disable `contacts`.

One thing I did not expect was that `python manage.py test` apparently runs tests even in disabled apps, which fail due to migrations not being executed. But that will not affect any deployment, just local development and I would not expect people to disable anything locally.

erichuanggit:
Looks good to me. 
But please `Don't` merge until next week. Let's not put more pressure on QA since we haven't planned this issue in this Sprint.

lao605:
@lubomir do we do all the migrations in deployment and then disable the apps? or we don't do testing on deployment env?

lao605:
@erichuanggit roger that!

lubomir:
@lao605 I believe if the apps are not in `INSTALLED_APPS`, their migrations are not run, but I have no idea what happens if the tables already exists when app is uninstalled. They might be removed or left unused. I don't think we need to actually test these scenarios.

PR #204, opened 2015-11-06
Exclude pdc-client from distribution package

simozhan:
JIRA: PDC-1094

Exclude the pdc_client module as it is packaged independently with tito


bliuredhat:
+1

PR #207, opened 2015-11-11
Add Component Contact client plugin

xychu:
Including:
  - list
  - info
  - create
  - delete
  - delete-match

Add 'DELETE' support in `test_helpers.py`;
Tests added.

JIRA: PDC-1164 & PDC-1165


lubomir:
Would it be possible to use [mutually exclusive group](https://docs.python.org/2/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) for `--username` and `--mail-name`? Right now it displays both as optional, and when you don't specify them, you get an error.

I'm not sure how useful the `info` command is. It displays no additional information that would not be present in the list view. To get the id, you need to run the list command anyway.

xychu:
@lubomir all cool advises, will update accordingly.

erichuanggit:
Looks good to me.

PR #208, opened 2015-11-11
Update Component client command to use new contact API.

xychu:
Will ignore old contacts if any.
JIRA: PDC-1166

### DIFF
--- a/pdc_client/plugins/component.py
+++ b/pdc_client/plugins/component.py
@@ -13,6 +13,15 @@ from pdc_client.plugin_helpers import (PDCClientPlugin,
                                        add_create_update_args)
 
 
+def update_component_contacts(component, component_contacts):
+    contacts = []
+    if component_contacts:
+        for contact in component_contacts:
+            contact.pop('component')
+            contacts.append(contact)
+        component['contacts'] = contacts
+
+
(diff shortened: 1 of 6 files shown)


pdc_client/plugins/component.py
```diff
@@ -70,6 +70,14 @@ def global_component_info(self, args, global_component_id=None):
         global_component_id = global_component_id or args.global_component_id
         global_component = self.client['global-components'][global_component_id]._()
 
+        component_contacts = get_paged(self.client['global-component-contacts']._, component=global_component['name'])
+        contacts = []
+        if component_contacts:
```
erichuanggit: This code looks like almost same as what you did for `release-component-contacts`. How about abstract it out to a single function avoiding code duplicated?

xychu: Sure, will update later.

lubomir:
Given that the logic in formatting contacts required a separate function, It seems like a good idea to have some tests for showing components with contacts.

xychu:
@lubomir good point and done.

erichuanggit:
Looks good to me.

lubomir:
:+1:

PR #209, opened 2015-11-12
pdc command take release and name as parameters instead of id.

ycheng-aa:
JIRA: PDC-1159

### DIFF
--- a/pdc_client/plugins/component.py
+++ b/pdc_client/plugins/component.py
@@ -4,6 +4,7 @@
 # Licensed under The MIT License (MIT)
 # http://opensource.org/licenses/MIT
 #
+import sys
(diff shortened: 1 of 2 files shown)


xychu:
:+1:

ycheng-aa:
The output of a exception looks not so friendly for user. How about print the error message and exit(1). Could it report the error by exiting code?

lubomir:
Printing the error and exitting would work as well. My main complaint is that the exit code should not be 0.

ycheng-aa:
Rebase to eliminate CI failed.

erichuanggit:
Looks good to me.

lubomir:
:+1:

PR #211, opened 2015-11-12
pdc global-component update should take its name instead of ID

bliuredhat:
When update the global-component, using GLOBAL_COMPONENT_NAME
instead of GLOBAL_COMPONENT_ID.

JIRA: PDC-1157

### DIFF
--- a/pdc/apps/compose/filters.py
+++ b/pdc/apps/compose/filters.py
@@ -10,7 +10,7 @@ from django.db.models import Q
 from django.core.exceptions import ValidationError
 
 from pdc.apps.common.filters import value_is_not_empty
-from .models import Compose, OverrideRPM
+from .models import Compose, OverrideRPM, ComposeTree
 
 
 class ComposeFilter(django_filters.FilterSet):
@@ -77,3 +77,15 @@ class OverrideRPMFilter(django_filters.FilterSet):
     class Meta:
         model = OverrideRPM
         fields = ('release', 'variant', 'arch', 'srpm_name', 'rpm_name', 'rpm_arch', 'comment')
+
+
+class ComposeTreeFilter(django_filters.FilterSet):
+    compose         = django_filters.CharFilter(name='compose__compose_id', lookup_type='iexact')
+    variant         = django_filters.CharFilter(name='variant__variant_uid', lookup_type='iexact')
+    arch            = django_filters.CharFilter(name='arch__name', lookup_type='iexact')
+    location        = django_filters.CharFilter(name='location__short', lookup_type='iexact')
+    scheme          = django_filters.CharFilter(name='scheme__name', lookup_type='iexact')
+
+    class Meta:
+        model = ComposeTree
+        fields = ('compose', 'variant', 'arch', 'location', 'scheme')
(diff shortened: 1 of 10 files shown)


pdc_client/plugins/component.py
```diff
@@ -66,6 +66,14 @@ def list_global_components(self, args):
                       global_component['id'],
                       global_component['name'])
 
+    def get_component_id(self, args):
```
erichuanggit: It's better to update this function's name -> `get_component_by_name`

pdc_client/plugins/component.py
```diff
@@ -66,6 +66,14 @@ def list_global_components(self, args):
...
+        global_component_name = args
```
erichuanggit: this variable looks like not necessary

pdc_client/tests/component/data/global_component/detail.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "1",
-    "name": "Test Global Component",
+    "name": "Test_Global_Component",
```
erichuanggit: Any benefit to update `Test Global Component` to `Test_Global_Component`?

bliuredhat: Because when run the command :" self.runner.run(['global-component', 'update', 'Test_Global_Component', '--name', 'new_test_name']),  " ;  
which corresponding to the pdc comand: "pdc  global-component  update  --name new_test_name  Test_Global_Component ". If global-component's name is "Test Global Component",  which exist blank space, the pdc command will split it to 3 parts,  and fail to filter the name.

lubomir: It's not `pdc` splitting it into 3 parts, that is done by Bash. If you want a single strings with spaces, quoting is required: `pdc global-component update --name "new test name" "Test Global Component"`.

bliuredhat:
@lubomir and @erichuanggit  ; Yes,  I will edit it.

pdc_client/plugins/component.py
bliuredhat: @erichuanggit , you are right~!

pdc_client/plugins/component.py
```diff
@@ -104,13 +112,17 @@ def global_component_create(self, args):
 
     def global_component_update(self, args):
         data = extract_arguments(args)
+        global_component_id = self.get_component_id(args.global_component_name)
+        if not global_component_id:
+            sys.stderr.write("This global component don't exist")
```
ycheng-aa: 'don't exist' should be 'doesn't exit'?
Also it's better add new line in sys.stderr.write.

pdc_client/plugins/component.py
```diff
@@ -64,6 +65,13 @@ def list_global_components(self, args):
                       global_component['id'],
                       global_component['name'])
 
+    def get_component_id(self, args):
```
ycheng-aa: Is it better name the function as _get_component_id? It's more like a class private function.

xychu:

bliuredhat:
I will close this branch because it have merged to pdc-client repo.

PR #214, opened 2015-11-13
Improve error handling in client

lubomir:
When network request fails, the error printing is more robust. Ideally,
the error response should be JSON encoded. In that case it will be
pretty printed.

If JSON parsing or pretty printing fails, user is prompted to file a
bug. The response data is logged only if it was valid JSON, as the only
case of non-JSON responses is a huge HTML dump.

The error is now consistently printed to stderr.

JIRA: PDC-1193

### DIFF
--- a/pdc_client/runner.py
+++ b/pdc_client/runner.py
@@ -4,6 +4,9 @@
 # Licensed under The MIT License (MIT)
 # http://opensource.org/licenses/MIT
 #
+
+from __future__ import print_function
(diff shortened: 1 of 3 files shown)


erichuanggit:
Looks good to me.

xychu:

lubomir:
Opened as https://github.com/product-definition-center/pdc-client/pull/5

PR #217, opened 2015-11-16
Add new endpoint API ComposeTree

simozhan:
JIRA: PDC-1100

### DIFF
--- a/pdc/apps/compose/models.py
+++ b/pdc/apps/compose/models.py
@@ -519,10 +519,10 @@ class Scheme(models.Model):
 class ComposeTree(models.Model):
     compose             = models.ForeignKey("Compose")
     variant             = models.ForeignKey("Variant")
-    arch                = models.CharField(max_length=50)
+    arch                = models.ForeignKey("common.Arch")
     location            = models.ForeignKey("Location")
     scheme              = models.ForeignKey("Scheme")
-    url                 = models.CharField(max_length=100)
+    url                 = models.CharField(max_length=255)
     synced_content      = models.ManyToManyField('repository.ContentCategory')
 
     class Meta:
@@ -537,7 +537,7 @@ class ComposeTree(models.Model):
         return {
             "compose": self.compose.compose_id,
             "variant": self.variant.variant_uid,
-            "arch": self.arch,
+            "arch": self.arch.name,
             "location": self.location.short,
             "scheme": self.scheme.name,
             "url": self.url,
--- a/pdc/apps/compose/views.py
+++ b/pdc/apps/compose/views.py
@@ -1417,13 +1417,13 @@ class ComposeTreeViewSet(ChangeSetModelMixin,
     -d _data_ (a json string). or GUI plugins for
     browsers, such as ``RESTClient``, ``RESTConsole``.
     """
-    queryset = ComposeTree.objects.all()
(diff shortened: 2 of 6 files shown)


xychu:
LGTM

pdc/apps/compose/views.py
```diff
@@ -1399,3 +1402,142 @@ def list(self, request):
...
+    """
+    API endpoint that allows querying compose-variant-arch relevant to location.
+
+    ##Test tools##
+
+    You can use ``curl`` in terminal, with -X _method_ (GET|POST|PUT|DELETE),
+    -d _data_ (a json string). or GUI plugins for
+    browsers, such as ``RESTClient``, ``RESTConsole``.
+    """
+    queryset = ComposeTree.objects.all()
```
ycheng-aa: If some other related objects also need to be accessed, maybe could use select_related() to fetch them with on DB hit.

pdc/apps/compose/views.py
```diff
@@ -1399,3 +1402,142 @@ def list(self, request):
...
+
+        updatable_keys = set(['scheme', 'url', 'synced_content'])
+        if set(request.data.keys()) - updatable_keys:
+            return Response(status=status.HTTP_400_BAD_REQUEST,
+                            data={'detail': 'Only these properties can be updated: %s'
+                                  % ', '.join(updatable_keys)})
+
+        return super(ComposeTreeViewSet, self).update(request, *args, **kwargs)
+
+    def bulk_update(self, *args, **kwargs):
```
ycheng-aa: Is it necessary to implement it? I think with BulkRouter, the default implementation is like this.

ycheng-aa:
Generally looks good to me.

pdc/apps/compose/models.py
```diff
@@ -505,3 +505,41 @@ class Meta:
...
+
+
+class Scheme(models.Model):
+    name = models.CharField(max_length=50, unique=True)
+
+
+class ComposeTree(models.Model):
+    compose             = models.ForeignKey("Compose")
+    variant             = models.ForeignKey("Variant")
+    arch                = models.CharField(max_length=50)
```
sochotnicky: I think these should be one of currently defined arches and not just any string.

pdc/apps/compose/models.py
```diff
@@ -505,3 +505,41 @@ class Meta:
...
+    location            = models.ForeignKey("Location")
+    scheme              = models.ForeignKey("Scheme")
+    url                 = models.CharField(max_length=100)
```
sochotnicky: max length 100 is definitely too restrictive. Why should the url be that short? We would certainly hit this limit

pdc/apps/compose/models.py
simozhan: Ok. fixed.

pdc/apps/compose/models.py
simozhan: Yes. it is too restrictive. change it to 255.

pdc/apps/compose/views.py
simozhan: Yes, the default implementation is like this. but the doc is not good enough, this api only support partial update.

lubomir:
Looks fine to me.

PR #218, opened 2015-11-17
Create script that can create release-components from compose informa…

ycheng-aa:
…tion

1. Take release from argument and find latest compose for it
2. For latest compose find all source RPMs
3. For each source RPM name create a release-component with identical
name in given release (if needed create global component with that name too)

JIRA: PDC-1133


lubomir:
Looks good to me.

simozhan:
Looks good to me

PR #220, opened 2015-11-18
Remove client documentation

lubomir:
The document should move to pdc-client repo. The autodoc extension that
was used in the document lead to errors when trying to import modules
that are not present in the repo now. Since there was no other use of
the extension, autodoc is disabled now.

The move should happen in https://github.com/product-definition-center/pdc-client/pull/4.


bliuredhat:
:+1:

simozhan:
:+1:

PR #222, opened 2015-11-19
Specify Accept instead of Content-Type in curl examples.

ralphbean:
I'm pretty sure this is what we want here.

Content-Type specifies what kind of data we're sending.. but we're not sending json here.

Accept specifies what kind of data you want back.. and we want JSON (not the html docs served at the same endpoint).

### DIFF
--- a/pdc/apps/auth/__init__.py
+++ b/pdc/apps/auth/__init__.py
@@ -128,7 +128,7 @@ Using Token
 -----------
 
     # obtain token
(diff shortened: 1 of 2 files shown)


lubomir:
You are completely correct. The documentation suggests the `Content-Type` header in a lot of places where it makes no sense (i.e. any `GET` request). On the other hand `Accept` is not mentioned at all, and it should be specified pretty much everywhere.

lubomir:
Merged as 53dbc406a56169bc8671618bbe72c18ac8ef0a7d.

PR #223, opened 2015-11-25
Use single string field for user names.

ycheng-aa:
User single string instead of first name and last
name combination.

JIRA: PDC-1051


bliuredhat:
+1

lubomir:
The migration seems needlessly complicated: why not add new field, copy full name, remove old fields?

ycheng-aa:
Good point. Fixed in d1dbad8ada10cb6. Thanks.

lubomir:
:+1:

xychu:
:+1:

PR #225, opened 2015-11-27
Drop old contact API.

simozhan:
Remove contacts from global and release component end-points.
Remove role contacts API.
Delete tests for those end-points.
Remove deprecation notices in modified resources.
Move all contact related viewsets, serializers, filters and tests to pdc.app.contact application.
Document and test bulk update on release components

JIRA: PDC-1050

### DIFF
--- a/pdc/apps/contact/migrations/0006_auto_20151127_0837.py
+++ b/pdc/apps/contact/migrations/0006_auto_20151127_0837.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contact', '0005_auto_20151027_0725'),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name='rolecontact',
+            unique_together=set([]),
+        ),
+        migrations.RemoveField(
+            model_name='rolecontact',
+            name='contact',
+        ),
+        migrations.RemoveField(
+            model_name='rolecontact',
+            name='contact_role',
+        ),
+        migrations.DeleteModel(
+            name='RoleContact',
+        ),
+    ]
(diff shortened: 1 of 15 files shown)


lubomir:
The role-contacts API should be deleted as well, I don't think it makes sense anymore.

simozhan:
@lubomir yes. remove it.

xychu:
:+1:

pdc/apps/contact/migrations/0006_auto_20151127_0837.py
```diff
@@ -0,0 +1,29 @@
...
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contact', '0005_auto_20151027_0725'),
+    ]
+
```
ycheng-aa: I think it also depends ('component', '0011_auto_20151126_0602'). If this migration runs before 0011, should it possibly fail because foreign key constraint?

pdc/apps/contact/migrations/0006_auto_20151127_0837.py
```diff
@@ -0,0 +1,29 @@
...
+        migrations.RemoveField(
+            model_name='rolecontact',
+            name='contact',
+        ),
+        migrations.RemoveField(
+            model_name='rolecontact',
+            name='contact_role',
+        ),
+        migrations.DeleteModel(
+            name='RoleContact',
```
ycheng-aa: Could we DeleteModel directly?

ycheng-aa:
Generally looks good tome. But I am not sure if migrations dependency could cause problem.

pdc/apps/contact/migrations/0006_auto_20151127_0837.py
simozhan: Good point. Actually this file is generated automatically by command "python manage.py makemigrations/migrate". Django framework will build a migration dependency graph using both the disk migration files and database. so i am assuming the framework will take care the dependency issue and it will not fail the migration.

pdc/apps/contact/migrations/0006_auto_20151127_0837.py
simozhan: This is generated automatically. But i tried to delete it and migrated again, it worked.

PR #234, opened 2015-12-07
A slightly more friendly default 404 response.

ralphbean:
When trying to import a compose, I kept getting a 404.

- It was hard to uncover the json error response from pdc-client,
- but once I got it, it was vague.  Just ``{'detail': 'Not found'}``.

It was only after I modified the pdc server code to include more
information, did I find it was because I needed to specify a missing
Arch entry.  Including the exception representation like this should
make it easier to solve problems like this in the future.


xychu:
Looks good to me.

simozhan:
:+1:

xychu:
Merged as 48ee825

PR #241, opened 2015-12-16
Add compose image RTT tests APIs.

xychu:
JIRA: PDC-1191

With the constraint that each compose in specific variant arch the image `file_name` instead of (`file_name`, `sha256`) is unique, we could use `compose/variant/arch/file_name` as lookup key.

### DIFF
--- a/pdc/apps/compose/views.py
+++ b/pdc/apps/compose/views.py
@@ -33,15 +33,17 @@ from pdc.apps.common.viewsets import (ChangeSetCreateModelMixin,
                                       NoEmptyPatchMixin,
                                       ChangeSetDestroyModelMixin,
                                       ChangeSetModelMixin,
-                                      MultiLookupFieldMixin)
+                                      MultiLookupFieldMixin,
+                                      ChangeSetUpdateModelMixin)
 from pdc.apps.release.models import Release
 from .models import (Compose, VariantArch, Variant, ComposeRPM, OverrideRPM,
                      ComposeImage, ComposeRPMMapping, ComposeAcceptanceTestingState,
                      ComposeTree)
 from .forms import (ComposeSearchForm, ComposeRPMSearchForm, ComposeImageSearchForm,
                     ComposeRPMDisableForm, OverrideRPMForm, VariantArchForm, OverrideRPMActionForm)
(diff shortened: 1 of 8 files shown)


pdc/apps/compose/views.py
```diff
@@ -1364,6 +1366,106 @@ def list(self, request, **kwargs):
...
+                                 mixins.RetrieveModelMixin,
+                                 StrictQueryParamMixin,
+                                 MultiLookupFieldMixin,
+                                 viewsets.GenericViewSet):
+    """
+    API endpoint that allows querying compose-image RTT Test results.
+
+    ##Test tools##
+
+    You can use ``curl`` in terminal, with -X _method_ (GET|POST|PUT|DELETE),
```
ycheng-aa: It looks DELETE is not supported here.

pdc/apps/compose/views.py
```diff
@@ -1364,6 +1366,106 @@ def list(self, request, **kwargs):
...
+        # This method is used by bulk update and partial update, but should not
+        # be called directly.
+        if not kwargs.get('partial', False):
+            return self.http_method_not_allowed(request, *args, **kwargs)
+
+        if not request.data:
+            return NoEmptyPatchMixin.make_response()
+
+        updatable_keys = set(['test_result'])
+        if set(request.data.keys()) - updatable_keys:
```
ycheng-aa: Should we just set fields to readonly=true in serializer except the updatable one then let drf framework to treat automatically?

ycheng-aa:
Generally looks good for me.

pdc/apps/compose/views.py
xychu: DRF default behavior is to ignore updating on readonly fileds not return a 400 Bad Request.
I know the hard coded set here is hard to trace and maintain, maybe we should find a better way to get rid of all these `updatable_keys` hacks?

pdc/apps/compose/views.py
xychu: Will fix it.

pdc/apps/compose/views.py
ycheng-aa: Then it's fine.

simozhan:
looks good to me

PR #243, opened 2015-12-21
Rpc/release/clone-components

bliuredhat:
This patch is used to clone release components from one  release to another release.
 The result will return target release.

JIRA: PDC-1204
sub JIRA: PDC-1252

### DIFF
--- a/pdc/apps/component/tests.py
+++ b/pdc/apps/component/tests.py
@@ -1175,6 +1175,31 @@ class ReleaseCloneWithComponentsTestCase(TestCaseWithChangeSetMixin, APITestCase
         self.assertIn('Value [foobar] of include_inactive is not a boolean',
                       response.content)
 
+    def test_release_components_clone(self):
+        args = {"name": "Supplementary", "short": "supp", "version": "1.1",
+                "release_type": "ga"}
+        target_response = self.client.post(reverse('release-list'), args)
+        self.assertEqual(target_response.status_code, status.HTTP_201_CREATED)
+        target_release_id = target_response.data['release_id']
+        response = self.client.post(reverse('releasecomponentclone-list'),
+                                    {'source_release_id': 'release-1.0',
+                                     'target_release_id': target_release_id,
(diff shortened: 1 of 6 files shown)


bliuredhat:
@lubomir ,@ycheng-aa,@simozhan,@

bliuredhat:
Hi,all; please review this patch~!

pdc/apps/component/tests.py
```diff
@@ -1175,6 +1175,19 @@ def test_clone_components_bad_include_inactive_value(self):
...
+        target_response = self.client.post(reverse('release-list'), args)
+        self.assertEqual(target_response.status_code, status.HTTP_201_CREATED)
+        target_release_id = target_response.data['release_id']
+        response = self.client.post(reverse('releasecomponentclone-list'),
+                                    {'source_release_id': 'release-1.0',
+                                     'target_release_id': target_release_id,
+                                     'component_dist_git_branch': 'new_branch'},
+                                    format='json')
+        self.assertEquals(response.status_code, status.HTTP_201_CREATED)
+
```
simozhan: I think more test cases should be added. For example,the test case for no releasecomponent in source release or there is inactive releasecomponent, etc.

pdc/apps/release/views.py
```diff
@@ -632,6 +632,64 @@ def create(self, request):
...
+        if source_release.releasecomponent_set.count() == 0:
+            return Response(target_data, status=status.HTTP_200_OK)
+
+        signals.rpc_release_clone_component.send(sender=target_release.__class__,
+                                                 request=request,
+                                                 original_release=source_release,
+                                                 release=target_release)
+
+        return Response(target_data, status=status.HTTP_201_CREATED)
+
```
simozhan: Why we return the target release instead of cloned release component?

bliuredhat: I think you are right , but I am not sure what value will return.

simozhan:
:+1:

PR #245, opened 2015-12-22
Handle ValueError in exception_handler

ycheng-aa:
When called compose-rpms api and rpm date info was
not correct, productmd would throw ValueError exception and
cause server internal error (500 error).
Now let exception_handler handle ValueError exception then server
will response http code 400 and meaningful message.


simozhan:
:+1:

ycheng-aa:
Rebased and merged.

PR #246, opened 2015-12-24
Fix the bug that build image rtt test resluts should not allow put me…

ycheng-aa:
…thod.


pdc/apps/package/views.py
```diff
@@ -476,11 +476,30 @@ def retrieve(self, request, *args, **kwargs):
         return super(BuildImageRTTTestsViewSet, self).retrieve(request, *args, **kwargs)
 
     def update(self, request, *args, **kwargs):
+        # This method is used by bulk partial update, but should not
```
simozhan: only for bulk partial update?

ycheng-aa: Thanks. Fixed in d92fccc2f0e27f,

ycheng-aa:
Reviewed the patch with @simozhan .

PR #247, opened 2015-12-31
Some improvements for stare image test results.

ycheng-aa:
1. Use compose id as compose out put, not compose model unicode.
   So the filter and output is consistent.
2. Response test_result illegal with default ChoiceSlugField behavior.
   It is more clear.

JIRA: PDC-1191


simozhan:
:+1:

bliuredhat:
:+1:

PR #248, opened 2016-01-04
Rpc/release/clone-components

bliuredhat:
Remove the 'include_trees' para in component clone. This ideal
get from lubomir; means that don't need to clone variant.

By the way, we don't need to create release-repo clone
in this task.

JIRA: PDC-1204
subTask: PDC-1052

### DIFF
--- a/pdc/apps/release/models.py
+++ b/pdc/apps/release/models.py
@@ -308,7 +308,6 @@ class Variant(models.Model):
         }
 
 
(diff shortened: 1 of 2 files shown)


bliuredhat:
@simozhan  and @ycheng-aa ; please review this patch , but I got some suggestion that the current format of result is not good, we need to discuss it in meeting.

PR #249, opened 2016-01-07
Add description for create method in compose-tree-locations api

bliuredhat:
JIRA: PDC-1278

### DIFF
--- a/pdc/apps/release/models.py
+++ b/pdc/apps/release/models.py
@@ -308,7 +308,6 @@ class Variant(models.Model):
         }
 
 
-@receiver(signals.rpc_release_clone_component)
 @receiver(signals.release_clone)
 def clone_variants(sender, request, original_release, release, **kwargs):
     """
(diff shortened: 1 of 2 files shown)


bliuredhat:
@simozhan  and @ycheng-aa ; please review this patch~!

simozhan:
Actually i am thinking this is not a bug(Sorry for not catching your meaning before). When you create a compose/variant/arch/location, compose/variant/arch combo should be already exist, if compose/variant/arch/location already exists, we don't need to create it, right?

bliuredhat:
@simozhan , yes, it's not bug; just the description is not clear.
And we all know; {compose}/{variant}/{arch}/{location} these 4 paras all are existed and the combo is unique.

simozhan:
@bliuredhat i think the description is ok as well. you don't need to emphasize that {compose}/{variant}/{arch}/{location} is unique, because they are unique together, if there is already the resource, framework will raise something like "* The fields compose, variant, arch, location must make a unique set.". if there is no location, framework will raise something like""* 'Test55' is not allowed value. Use one of 'NAY', 'BRQ', 'PEK', 'Test1', 'Test2'. if there is no combination for "{compose}/{variant}/{arch}", our code will raise error "The combination with compose %s, variant %s, arch %s does not exist". So the description is to tell the user, the combination must exist, not only compose, variant, arch exists independently.

PR #250, opened 2016-01-07
Fix cmposes/{compose_id}/rpm-mapping/{package}] return error if 'acti…

ycheng-aa:
…on' is invalid

Fix cmposes/{compose_id}/rpm-mapping/{package}] return error if
'action' value is invalid.

JIRA: 1271

### DIFF
--- a/pdc/apps/compose/models.py
+++ b/pdc/apps/compose/models.py
@@ -500,6 +500,8 @@ class OverrideRPM(models.Model):
                 orpm.include = not data['include']
                 orpm.save()
                 new_val = orpm.export()
+        else:
+            raise ValueError("action should only be 'create' or 'delete'")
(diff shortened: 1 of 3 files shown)


simozhan:
it should be better if there is a test case.

ycheng-aa:
@simozhan Thanks for the careful review. This is the bug we discussed in planning meeting and we agreed test is not necessary for this one. Also estimated it according to it.

simozhan:
maybe i misunderstood something, but if no test case, this case will not be covered and our coverage will decrease, right

ycheng-aa:
Sounds reasonable. Added test in f0df0f7a15fb2c4. Thanks.

simozhan:
:+1:

bliuredhat:
:+1:

PR #251, opened 2016-01-11
Add error info when input error para

bliuredhat:
This patch will add error info when Source_release or Target_release
is not existed. And then, update the decription in webUI.

JIRA:PDC-1204


bliuredhat:
@ycheng-aa ,@simo, @lubomir ; please help to review this patch.

lubomir:
Generally looks good. I would maybe change the error messages to actually include the incorrect value: "Source release %s does not exist." or something like that.

bliuredhat:
@lubomir ; thank you; your suggestion is good~!

simozhan:
Good for me

PR #252, opened 2016-01-11
Try to improve composes list performance.

ycheng-aa:
1. Get compose to Sigkeys' key_id mapping wholly instead of
   one by one.
2. Use select_related in some places to reduce DB access.

JIRA: PDC-1254

### DIFF
--- a/pdc/apps/compose/models.py
+++ b/pdc/apps/compose/models.py
@@ -87,10 +87,14 @@ class Compose(models.Model):
         """
         sigkeys = common_models.SigKey.objects.filter(composerpm__variant_arch__variant__compose=self).distinct()
         result = sorted([i.key_id for i in sigkeys])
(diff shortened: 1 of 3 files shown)


simozhan:
Is it ok to use https://docs.djangoproject.com/en/1.9/ref/utils/#django.utils.functional.cached_property
on def sigkeys(self):
Actually i haven't tried it whether it can improve performance or not, just want to give more ways.

ycheng-aa:
Thanks a lot. It is a pretty useful trick. For sigkeys cache, I found get sigkey once for a compose is very expensive. I just tried to fetch all  compose-sigkeys mapping in several DB queries and cache it. For this particular purpose, maybe don't have to use  cached_property here. This is definitely useful new knowledge for me. I will check again if I can use it in this task. Thanks.

simozhan:
Ok. The point is you cache all compose-sigkeys mapping. That sounds reasonable. :+1:

bliuredhat:
:+1:

PR #255, opened 2016-01-14
Improve output when successfully import files composeinfo.json/rpm-manifest.json/image-manifest.json

simozhan:
1. Return count of the images/rpms when importing image-manifes and rpm-manifest.
2. Return url when importing release from composeinfo.json

JIRA: PDC-1268

### DIFF
--- a/pdc/apps/compose/lib.py
+++ b/pdc/apps/compose/lib.py
@@ -169,6 +169,8 @@ def compose__import_rpms(request, release_id, composeinfo, rpm_manifest):
                               'num_linked_rpms': imported_rpms,
                           }))
 
+    return compose_obj.compose_id, imported_rpms
+
 
 @transaction.atomic
 def compose__import_images(request, release_id, composeinfo, image_manifest):
@@ -249,6 +251,8 @@ def compose__import_images(request, release_id, composeinfo, image_manifest):
                               'compose': compose_obj.compose_id,
                               'num_linked_images': imported_images,
                           }))
+
+    return compose_obj.compose_id, imported_images
 
 
 def _find_composes_srpm_name_with_rpm_nvr(nvr):
--- a/pdc/apps/compose/tests.py
+++ b/pdc/apps/compose/tests.py
@@ -734,12 +734,23 @@ class ComposeRPMViewAPITestCase(TestCaseWithChangeSetMixin, APITestCase):
                                      'composeinfo': self.compose_info},
                                     format='json')
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data.get('compose'), 'TP-1.0-20150310.0')
+        self.assertEqual(response.data.get('imported rpms'), 6)
         self.assertNumChanges([11, 5])
         self.assertEqual(models.ComposeRPM.objects.count(), 6)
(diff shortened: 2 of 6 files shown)


pdc/apps/compose/lib.py
```diff
@@ -55,6 +56,9 @@ def insert_compose_rpms_if_nonexist(compose_rpms_in_db, cursor,
                                       sigkey_id,
                                       path_id)
         compose_rpms_in_db.add(key)
+        inserted = True
+
+    return inserted
```
ycheng-aa: How about here return False and line 59 return True, then don't have to use inserted.

pdc/apps/compose/lib.py
```diff
@@ -149,25 +153,29 @@ def compose__import_rpms(request, release_id, composeinfo, rpm_manifest):
...
                     content_category = rpm_data["category"]
                     content_category_id = repository_models.ContentCategory.get_cached_id(content_category)
-                    insert_compose_rpms_if_nonexist(compose_rpms_in_db, cursor,
-                                                    var_arch_obj.id, rpm_id,
-                                                    content_category_id, sigkey_id, path_id)
+                    inserted = insert_compose_rpms_if_nonexist(compose_rpms_in_db, cursor,
+                                                               var_arch_obj.id, rpm_id,
+                                                               content_category_id, sigkey_id, path_id)
+                    if inserted:
+                        imported_rpms += 1
```
ycheng-aa: Minor thing. Maybe it could 
if insert_compose_rpms_if_nonexist(compose_rpms_in_db, cursor,
                                                              var_arch_obj.id, rpm_id,
                                                               content_category_id, sigkey_id, path_id):
     imported_rpms += 1

ycheng-aa:
Generally looks good for me. A few minor things.

bliuredhat:
:+1

lubomir:
The improved responses look good to me. I would maybe even simplify it and always return the same object, even with 0 imported things.

simozhan:
@lubomir do you mean if the user imported the same file the second time, we return the same result as the first time. for example {"compose_id": "RHELXXXX", "imported rpms": 68330}. i am happy with this as well. how about the status code? HTTP_201_CREATED for all or HTTP_201_CREATED for first time, HTTP_200_OK for second time. i don't have strong opinion. Any one should be ok for me

simozhan:
@lubomir i think this patch is simpler. i bring back the notice changlog because i think it is reasonable to add a "notice" when importing rpms or images even though no real rpms imported.

lubomir:
I think the correct counting is good.

What I meant was that in the viewset, the response should always contain the object `{"compose": "XXX", "imported rpms": Y}` even if Y=0. This is for the user who called the import, so that they know what happened.

In the changelog, inserting a notice that says nothing happened does not seem that useful, so I would prefer previous version, i.e. only insert notice if count > 0.

I'm sorry for the confusion.

simozhan:
It is ok. no confusion for me :)
I am just thinking maybe the user will think it is weird if they can get response {"compose": "XXX", "imported rpms": Y}, but no changelog. so i keep the notice changelog, although it doesn't include other meaningful changes

lubomir:
Ok then. :+1:

PR #258, opened 2016-01-18
Add response info for composes/{compose_id}/rpm-mapping/{package}

bliuredhat:
JIRA:PDC-1257


simozhan:
:+1:

pdc/apps/compose/views.py
```diff
@@ -813,6 +813,17 @@ def update(self, request, **kwargs):
         """
         __URL__: $LINK:composerpmmapping-detail:compose_id:package$
 
+        __Data__:
+            {
+                Variants:{
+                    archs:{
```
ycheng-aa: It looks like the indent is not very inconsistent with other APIs.

PR #261, opened 2016-01-20
Provide a response header field name "PDC-Warning"

ycheng-aa:
Warning could be put into this field. Currently, it is
added into the response from NoEmptyPatchMixin when
patch or update with empty data.

JIRA: PDC-1283

### DIFF
--- a/pdc/apps/common/constants.py
+++ b/pdc/apps/common/constants.py
@@ -18,3 +18,6 @@ DEVEL_CONTACT = 'Devel_Owner'
 
 # Arch type
 ARCH_SRC = 'src'
+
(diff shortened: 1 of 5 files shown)


simozhan:
looks good to me.

ycheng-aa:
I will change the PDC-Warning to pdc-warning in commit message title in final rebase.

lubomir:
I think this is good. I have no preference over lower- or upper-case, as HTTP headers are case insensitive ([RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)).

simozhan:
:+1:

PR #262, opened 2016-01-21
Add optional "built-for-release" field in rpms resource

ycheng-aa:
JIRA: PDC-818

### DIFF
--- a/pdc/apps/package/models.py
+++ b/pdc/apps/package/models.py
@@ -27,6 +27,7 @@ class RPM(models.Model):
     release             = models.CharField(max_length=200, db_index=True)
     arch                = models.CharField(max_length=200, db_index=True)  # nosrc
     srpm_name           = models.CharField(max_length=200, db_index=True)  # package (name of srpm)
+    built_for_release   = models.ForeignKey('release.Release', null=True, blank=True)
     srpm_nevra          = models.CharField(max_length=200, null=True, blank=True, db_index=True)
     # Well behaved filenames are unique, but that is enforced by having unique NVRA.
     filename            = models.CharField(max_length=4096)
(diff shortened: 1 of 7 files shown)


simozhan:
Generally looks good to me. just some naming maybe we can't export to outside.

pdc/apps/package/models.py
```diff
@@ -27,6 +27,7 @@ class RPM(models.Model):
     release             = models.CharField(max_length=200, db_index=True)
     arch                = models.CharField(max_length=200, db_index=True)  # nosrc
     srpm_name           = models.CharField(max_length=200, db_index=True)  # package (name of srpm)
+    koji_build_target   = models.CharField(max_length=200, null=True, blank=True, db_index=True)
```
dmach: I don't think this field belongs here.
PDC should be about *metadata* storage.
It shouldn't refer to any systems or services (with the exception of bindings).
I propose to move this field outside RPM model to some external module or extension and map it via OneToOne relation.

dmach: Actually, if you want to track "RPM was built in a release" relation,
why don't you add a RPM.built_for_release = ForeignKey(Release, null=True, blank=True) field?

That's something completely unrelated to infrastructure and it's up to an external script / user to provide the correct data.

ycheng-aa: As @sochotnicky explained in JIRA, it looks can't use built_for_release. If one koji build target is corresponding to one rpm and currently is not used by other models, it is much simplier just add a char field. 
Even move this field outside RPM model to some external module, is it still in PDC? If so, I don't quite understand why this follows the "PDC should be about metadata storage" rule.

dmach: What if I want to use PDC without koji? How can I map builds to releases then?

Or how to deal with several koji instances?
For example Fedora uses primary koji and secondary-arch koji.
Then this field is useless, because PDC knows nothing about koji instances and this field refers to generic koji.

I'd rather create a relation and let someone/something outside PDC decide how builds map to releases.

ycheng-aa: RPM has linked_releases  
 linked_releases     = models.ManyToManyField('release.Release', related_name='linked_rpms')
Is it what you want?

ycheng-aa: This task's AC has been changed. I will follow the new AC to implement it. Thanks for the discussion.

bliuredhat:
looks good for me~:+1:

simozhan:
looks good for me

PR #263, opened 2016-01-25
Return error info When srouce release doesn't contain component

bliuredhat:
For rpc/release/component-clone, when source-release doesn't contain
component, return error info for user.

JIRA: PDC-1282

### DIFF
--- a/pdc/apps/release/views.py
+++ b/pdc/apps/release/views.py
@@ -666,8 +666,8 @@ class ReleaseComponentCloneViewSet(StrictQueryParamMixin, viewsets.GenericViewSe
             "url": [the link for target release component]
             }
 
-        If the source-release doesn't contains components, which will return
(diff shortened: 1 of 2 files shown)


bliuredhat:
@ycheng-aa and @simozhan ; please help to review this patch~thank you!

pdc/apps/release/views.py
```diff
@@ -659,11 +659,16 @@ def create(self, request):
...
             }
+
         __Response__:
+
             {
             "url": [the link for target release component]
             }
 
+        If the source-release doesn't contains components, which will return
+        400  {'detail': 'there is no component in source release'}.
```
simozhan: Is it better to change the description to
"Please make sure source release contains components, cloning source release without components doesn't make much sense."

ycheng-aa:
Please also add test case.

PR #264, opened 2016-01-25
Add support for regexp for contact searches by component name

ycheng-aa:
User could filter release component contacts or golbal component
contacts by component name with regexp.
The implementation is based on Django query set regexp implementation.
So they should have the same limitations.

JIRA: PDC-1285

### DIFF
--- a/pdc/apps/common/filters.py
+++ b/pdc/apps/common/filters.py
@@ -8,6 +8,7 @@ import functools
 
 from django import forms
 from django.core.validators import EMPTY_VALUES
+from django.db.models import Q
 from django.utils import six
 from django_filters.filterset import (BaseFilterSet,
                                       FilterSet,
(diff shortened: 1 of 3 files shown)


bliuredhat:
:+1:

pdc/apps/common/filters.py
```diff
@@ -50,6 +52,23 @@ def filter(self, qs, value):
         return qs
 
 
+class MultiValueRegexFilter(MultiValueFilter):
```
simozhan: It seems the query operation in DRF supports &(And), will this change it to OR?
Why we want to support multi value regex?

ycheng-aa: For this task to filter contacts by component, previous implementation is MultiValueFilter. I think maybe it's better to support multi-value as before so not to break some existing script. The MultiValueFilter is using 'in', so I think the relationship is 'or' for multi-value.

simozhan: yes. you are right. i am just thinking for different field, for example, component and role, & means "AND", but for component and component, & means "OR". i am just thinking if it is confused for the end user. Maybe this is my unnecessary concern. :)

simozhan:
:+1:

PR #267, opened 2016-01-26
Add error message for compose-tree-locations.

bliuredhat:
When input error Compose or Variant, return error info.

JIRA:PDC-1232

### DIFF
--- a/pdc/apps/compose/tests.py
+++ b/pdc/apps/compose/tests.py
@@ -1973,6 +1973,27 @@ class ComposeTreeAPITestCase(TestCaseWithChangeSetMixin, APITestCase):
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         self.assertNumChanges([1])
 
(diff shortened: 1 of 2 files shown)


bliuredhat:
@simozhan; I have gotten suggestion from you about this task and  your ideal is reasonable and available. But I also check other api how to deal with the similar problem, such as Release/Component/Package  apis;  more likely to analysis the data in view.py file, so I consider whether we can continue to edit it  as current status?

simozhan:
@bliuredhat i haven't checked the cases one by one, maybe some case needs to check data in view, maybe some can be refactored. My point is the serializer is helped to encode/decode the data. it is better to put the data validation in serializer as much as possible and make the view clean. 
By the way, why to try/except get_object_or_404, to customize the error message?

bliuredhat:
@simozhan, thank you ; and then, QE have a requirement that the error message shall be clearly for user, so customize the error message with try/except.

PR #268, opened 2016-01-27
Remove the unavailable code

bliuredhat:
As JIRA shows,the update method of release-component have
updated,  and this task don't need to do.

JIRA: PDC-945


ycheng-aa:
Code looks good. I think the commit message could be more specific. After long time, we could still know the backgroud from commit message.

bliuredhat:
@ycheng-aa; you suggestion is to the point.

simozhan:
looks good to me

ycheng-aa:
Looks good.　I still think 'As JIRA shows' not that necessary. It's better to provide information in commit message itself.

bliuredhat:
@ycheng-aa; I got it; thank you~!

PR #269, opened 2016-01-28
Make MultiValueRegexFilter could treat empty string or wrong regexp f…

ycheng-aa:
…ormat

JIRA: PDC-1285

### DIFF
--- a/pdc/apps/common/filters.py
+++ b/pdc/apps/common/filters.py
@@ -7,6 +7,7 @@
 import functools
 
 from django import forms
+from django.core.exceptions import ValidationError
 from django.core.validators import EMPTY_VALUES
 from django.db.models import Q
 from django.utils import six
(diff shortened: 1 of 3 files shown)


simozhan:
I tried https://pdc.engineering.redhat.com/rest_api/v1/releases/?release_id=%28%28
and https://pdc.engineering.redhat.com/rest_api/v1/releases/?release_id=
it seems the first one returns nothing and the second one returns all releases. 
This implementation changes the behaviour, i am just wondering why not to keep the behaviour.

ycheng-aa:
@simozhan  I think it's because release_id in ReleaseFilter is not MultiValueFilter, and this filter is followed MultiValueFilter behavior. In MultiValueFilter it seems is this kind of behavior for release-component-contacts/?component=.  Please see the expected result part in https://bugzilla.redhat.com/show_bug.cgi?id=1302550

simozhan:
@ycheng-aa so then the problem is for different api, the search behavior is different. i think we should make them consistent.

ycheng-aa:
@simozhan Yes, it looks more reasonable. For this task itself I think just support regexp and follow previous behavior, just like QA expected. Make them consistent may should be in a separate task. But I am not sure if this kind of change will break some existing script.

simozhan:
it looks good to me for this task and i am ok if there is another ticket to track the issue.

bliuredhat:
looks good for me

PR #271, opened 2016-02-01
Update the error info for compose-tree-location

bliuredhat:
When variant para is missing or unavailable, return clear error info for user. 

JIRA: PDC-1232

### DIFF
--- a/pdc/apps/compose/serializers.py
+++ b/pdc/apps/compose/serializers.py
@@ -113,10 +113,10 @@ class ComposeTreeVariantField(serializers.Field):
 
     def to_internal_value(self, data):
         request_data = self.context.get("request").data
-        compose = None
-        if request_data:
-            compose = request_data.get("compose")
-        variant = Variant.objects.get(compose__compose_id=compose, variant_uid=data)
+        try:
+            variant = Variant.objects.get(compose__compose_id=request_data['compose'], variant_uid=data)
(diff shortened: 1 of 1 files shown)


lubomir:
How about this message: "Variant %s does not exist in compose %s."

bliuredhat:
@lubomir , thank you~!

bliuredhat:
@ycheng-aa  and @simozhan ; please review this patch~!

simozhan:
looks good

pdc/apps/compose/serializers.py
```diff
@@ -113,10 +113,14 @@ class ComposeTreeVariantField(serializers.Field):
 
     def to_internal_value(self, data):
         request_data = self.context.get("request").data
-        compose = None
-        if request_data:
-            compose = request_data.get("compose")
-        variant = Variant.objects.get(compose__compose_id=compose, variant_uid=data)
+        variant_all = Variant.objects.all()
+        if data not in [tmp['variant_uid'] for tmp in variant_all.values()]:
```
bliuredhat: @simozhan , I think this step can been removed, we don't need to verify the variant whether exist in DB or unrelated compose.

simozhan: I am ok to delete these lines. Only check if variant in related compose  is enough.

PR #274, opened 2016-02-09
Remove productmd hacks, deserialize data directly.

dmach:


lubomir:
Looks good to me. As discussed, doing the same thing with context manager would [look like this](https://gist.github.com/lubomir/721dd62b48c2bcf84443), but I don't think it is much better.

ycheng-aa:
Looks good.

simozhan:
:+1:

lubomir:
Merged as 8d24893fffa.

PR #275, opened 2016-02-09
Improve ordering releases, product versions and products.

dmach:
All objects were ordered by ID which was confusing for WebUI users.
Everything is now ordered by pretty IDs (e.g. release_id).


lubomir:
Looks good to me. Once we get into double digit versions, it will stop looking so nice (`foo-1` < `foo-10` < `foo-2`), but overall it's still an improvement.

ycheng-aa:
Looks good.

simozhan:
:+1:

lubomir:
Merged as b975e5d114572be24.

PR #276, opened 2016-02-17
Add more detail error info for compose-tree-locations

bliuredhat:
When Arch para exist in data but doesn't exist in given
compose/variant branch, return detail error info.

JIRA: PDC-1232


simozhan:
The fix is ok for me. It is better if there is a test case for it.

bliuredhat:
@simozhan , there is a test case for this patch. Located : pdc.apps.compose.tests.ComposeTreeAPITestCase.test_create_composetree_with_wrong_arch .

simozhan:
@bliuredhat the case is ok. please change the logic a bit as we discussed.

PR #278, opened 2016-02-18
Remove permissions from inactive user accounts

ycheng-aa:
If a member is staff but not in any group other than
user group itself, set the user to inactive and remove
from all groups in PDC.

JIRA: PDC-1345

### DIFF
--- a/pdc/pdc-sync-ldap
+++ b/pdc/pdc-sync-ldap
@@ -11,6 +11,9 @@ import django
 from optparse import OptionParser
 
 
+SERVICE_NAME_SEPARATOR = '/'
+
+
 def process_raw_ldap_data(groups):
     """
     Take data from LDAP and return mapping from group names to list of users.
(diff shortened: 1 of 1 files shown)


lubomir:
Is it actually necessary to create a set of all active users? This could get really large.

Would it work the same if we just assumed that a user is inactive if it's not a service and LDAP does not return anything?

ycheng-aa:
I am not sure about the meaning of *LDAP does not return anything* here. It looks LDAP will return the information of a group name and a list of corresponding users. Do you mean we can know directly whether a user is in a group except the user group itself?

lubomir:
I meant this condition:
```
data = backends.get_ldap_user(conn, user.username)
if not data:
    # user not in LDAP
```

ycheng-aa:
Fixed in 0771dc59b76. Thanks.

bliuredhat:
looks good for me

simozhan:
looks good

PR #282, opened 2016-02-22
variant types rename

ycheng-aa:


bliuredhat:
looks good for me; but I find this api without unit test; do we need to add test case for it?

simozhan:
looks good to me

ycheng-aa:
@bliuredhat This task is just renaming, and existing code don't have UT. I think the reason maybe it totally use its super classes' implementation with no change at all.

PR #296, opened 2016-03-06
Display deleted compose with special style in web UI.

ycheng-aa:
There will be a line cross the deleted compose and
text will be faded out in web UI.

JIRA: PDC-1347

### DIFF
--- a/pdc/apps/compose/templates/compose_list_include.html
+++ b/pdc/apps/compose/templates/compose_list_include.html
@@ -14,7 +14,11 @@
  </thead>
  <tbody>
 {% for compose in compose_list %}
-  <tr>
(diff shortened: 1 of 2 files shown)


lubomir:
:+1: I think the CSS should also be applied internally, as the style sheets are not shared in any way.

ycheng-aa:
@lubomir  Is that mean this style sheets should be defined in compose_list_include.html internally?

lubomir:
No. The style should be defined here. It should also be in rhpdc internally. This PR is good to be merged without any changes.

ycheng-aa:
Thanks for pointing out. I will do it in internal code.

PR #297, opened 2016-03-08
Fix the full compose import API web page footer error.

ycheng-aa:
Add queryset to pass permissions check to
avoid ;server internal error' in web page footer.

JIRA: PDC-1393


bliuredhat:
looks good for me

simozhan:
:+1:

PR #299, opened 2016-03-08
Add subvariant field to images

lubomir:
Productmd has merged a patch that add an extra field `subvariant` to image manifest. This in turn broke PDC test suite. This patch should update PDC to be able to store and return that value.


ralphbean:
@ausil is going to start using the patched productmd in Fedora releases today, so I'm going to need to build this patch into a new PDC release for Fedora.

@lubomir, can we move ahead with review here?  Or are we still blocked on something?

ralphbean:
We're going to try and roll this out in python-pdc-0.3.rc-2.git.54.g3849431.fc23.src.rpm, http://koji.fedoraproject.org/koji/taskinfo?taskID=13312863

ralphbean:
This is live now.  :)

PR #300, opened 2016-03-09
Indicate a compose is deleted in its detail page.

ycheng-aa:
Add special format and text in deleted compose detail page
heading part to indicate this compose is already deleted.

JIRA: PDC-1347


simozhan:
looks good to me.

PR #307, opened 2016-03-16
Productmd changes

dmach:
Added release_type to BaseProduct.
Validation of Release, ProductVersion, Product and BaseProduct fields is based on regexps from productmd.

Change release_ids to use '@' as separator between release and base product part.
Release_type can be in both parts of release_id now (for example foo-1.0-updates@fedora-23-updates).

### DIFF
--- a/pdc/apps/release/migrations/0005_auto_20160314_1532.py
+++ b/pdc/apps/release/migrations/0005_auto_20160314_1532.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from productmd.common import create_release_id
+
+
+def set_baseproduct_id(obj):
+    obj.base_product_id = create_release_id(obj.short.lower(), obj.version, obj.release_type.short)
+
+
+def set_release_id(obj):
+    bp_dict = {}
+    if obj.base_product:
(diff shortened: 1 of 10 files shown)


pdc/apps/release/migrations/0005_auto_20160314_1532.py
```diff
@@ -0,0 +1,45 @@
...
+    for i in model.objects.all():
+        i.type = release_type
+        i.save()
+
+
+def create_new_release_id(apps, schema_editor):
+    model = apps.get_model('release', 'Release')
+    for i in model.objects.all():
+        # this triggers populate_release_id() on pre-save and agenerate new release_id
+        i.save()
```
lubomir: This is not correct: according to [the documentation](https://docs.djangoproject.com/en/1.9/topics/migrations/#historical-models), the migration does not have access to any custom methods, so it does not trigger regeneration of the field. It would be better to duplicate code from `get_release_id` method here to make sure the migration is always using that.

dmach:
Release ID is explicitly re-generated now.

dmach:
Migration no longer depends on a model method.

ycheng-aa:
Looks good.

simozhan:
looks good to me

lubomir:
:+1:

PR #308, opened 2016-03-16
Fixes to fedmsg message publication.

ralphbean:
We hit this in production in Fedora just a few minutes ago.  The 'topic' here
in the PDC code was ``".compose"``.  Once that made it through the pipes it
came out the other side as ``'org.fedoraproject.prod.pdc..compose'`` (the
double dots in the middle are problematic).  This patch should take care of it
in the future.  (Although, perhaps you want to make the change somewhere else
more general that will help out the other messaging plugins as well?)

Furthermore, the ``msg`` was already encoded as JSON (which fedmsg then re-encoded).  So, we decode in the send_message method to avoid that.  (Example:  https://apps.fedoraproject.org/datagrepper/id?id=2016-83d610b3-da7a-4a44-b5f0-9edf75e795a4&is_raw=true&size=extra-large )

### DIFF
--- a/pdc/apps/utils/messaging.py
+++ b/pdc/apps/utils/messaging.py
@@ -36,7 +36,7 @@ class FedmsgMessenger(object):
         self.messenger = fedmsg
 
     def send_message(self, topic, msg):
-        self.messenger.publish(topic=topic, msg=msg)
+        self.messenger.publish(topic=topic.strip('.'), msg=msg)
 
 
 class ProtonMessenger(object):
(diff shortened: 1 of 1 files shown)


lubomir:
Apart from the missing import of `json` in `messaging.py` this looks good to me.

ralphbean:
Doh.  Missing import added in 3c5bbff (committed yesterday, but pushed just now).

lubomir:
I have squashed the fixup commit and merged. Thank you.

PR #313, opened 2016-03-21
Compose-tree-locations be unique over scheme

bliuredhat:
When with same compose/variant/arch/location and diff scheme, it also can create a instance for Compose-tree-locations.

JIRA: PDC-1389


simozhan:
looks good to me.

ycheng-aa:
:+1:

PR #314, opened 2016-03-21
Send signal when product is created or updated.

ycheng-aa:
So binding part could customize its behavior accroding
to signals and the data with the signal.
For example, record object data before and after update.

JIRA: PDC-1321


simozhan:
looks good to me

bliuredhat:
looks good for me

PR #315, opened 2016-03-21
Compose import images/rpms and full import APIs extra parameter not a…

ycheng-aa:
…llowed.

In compose import images/rpms and full import APIs, extra
parameter is not allowed.
Add tests for these APIs and compose tree location API.

JIRA: PDC-1403


simozhan:
looks good to me

bliuredhat:
looks good for me

PR #318, opened 2016-03-22
Visible for Arch names on compose page

bliuredhat:
When the value of Arch with a long string, align left.

JIRA: PDC-1226

### DIFF
--- a/pdc/apps/compose/templates/compose_detail.html
+++ b/pdc/apps/compose/templates/compose_detail.html
@@ -89,6 +89,7 @@
 .compose-links.dl-horizontal dt {
     width: 4em;
     overflow: visible;
+    text-align: right;
     text-overflow: clip;
     white-space: nowrap;
 }


pdc/apps/compose/templates/compose_detail.html
```diff
@@ -89,6 +89,7 @@ <h3 class="sub-header">{% trans 'Variants' %}</h3>
 .compose-links.dl-horizontal dt {
     width: 4em;
     overflow: visible;
+    text-align: right;
```
ycheng-aa: The commit message is `When the value of Arch with a long string, align right.`. Does here mean it's always align right?

simozhan:
:+1:

pdc/apps/compose/templates/compose_detail.html
bliuredhat: yes; align right~!

ycheng-aa: So maybe commit message should be changed a little.

bliuredhat:
@ycheng-aa  and @simozhan ;thank you~!

PR #324, opened 2016-03-23
Check compose rpm mapping parameters strictly and fix bug.

ycheng-aa:
1. For partial update, 'include' is only for 'action' is 'create'.
2. If 'action' is 'delete', treat include value properly in
   model.
3. Check parameters format in PUT method.

JIRA: PDC-1335, PDC-1390


bliuredhat:
looks good for me

simozhan:
:+1:

PR #325, opened 2016-03-23
Create Rpc/Overrides-rpm/clone.

bliuredhat:
Clone overrides-rpm from one release to annother.

Each optional argument specifies which type of overrides get copied.
If multiple are specified all have to match for override to be copied.

Both releases have to exist.

If target_release already has overrides, they do not get removed.

JIRA: PDC-1343

### DIFF
--- a/pdc/apps/compose/tests.py
+++ b/pdc/apps/compose/tests.py
@@ -730,6 +730,98 @@ class OverridesRPMAPITestCase(TestCaseWithChangeSetMixin, APITestCase):
         self.assertEqual(models.OverrideRPM.objects.count(), 0)
 
 
+class OverridesRPMCloneAPITestCase(TestCaseWithChangeSetMixin, APITestCase):
+    fixtures = [
+        'pdc/apps/release/fixtures/tests/release.json',
+        'pdc/apps/compose/fixtures/tests/compose_overriderpm.json',
+    ]
+
+    def setUp(self):
+        self.release = release_models.Release.objects.get(release_id='release-1.0')
+        self.override_rpm = {'id': 1, 'release': 'release-1.0', 'variant': 'Server', 'arch': 'x86_64',
+                             'srpm_name': 'bash', 'rpm_name': 'bash-doc', 'rpm_arch': 'x86_64',
+                             'include': False, 'comment': '', 'do_not_delete': False}
+
+    def test_clone_overridesRPM(self):
+        args = {"name": "Supplementary", "short": "supp", "version": "1.1",
+                "release_type": "ga"}
+        target_response = self.client.post(reverse('release-list'), args)
+        self.assertEqual(target_response.status_code, status.HTTP_201_CREATED)
+        target_release_id = target_response.data['release_id']
+        response = self.client.post(reverse('overridesrpmclone-list'),
+                                    {'source_release_id': 'release-1.0',
+                                     'target_release_id': target_release_id},
+                                    format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    def test_clone__overridesRPM_with_filter_rpm_name(self):
+        args = {"name": "Supplementary", "short": "supp", "version": "1.1",
+                "release_type": "ga"}
+        target_response = self.client.post(reverse('release-list'), args)
+        target_release_id = target_response.data['release_id']
+        self.override_rpm["rpm_name"] = "bash-debuginfo"
+        del self.override_rpm["id"]
+        response1 = self.client.post(reverse('overridesrpm-list'), self.override_rpm)
+        self.assertEqual(response1.status_code, status.HTTP_201_CREATED)
+        response = self.client.post(reverse('overridesrpmclone-list'),
+                                    {'source_release_id': 'release-1.0',
+                                     'target_release_id': target_release_id,
+                                     'rpm_name': "bash-debuginfo"},
+                                    format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    def test_clone_overridesRPM_with_error_target_release(self):
+        response = self.client.post(reverse('overridesrpmclone-list'),
+                                    {'source_release_id': 'release-1.0',
+                                     'target_release_id': 'release-2.0'},
+                                    format='json')
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+
+    def test_clone_overridesRPM_without_target_release(self):
+        response = self.client.post(reverse('overridesrpmclone-list'),
+                                    {'source_release_id': 'release-1.0'},
+                                    format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_clone_overridesRPM_with_error_source_release(self):
+        response = self.client.post(reverse('overridesrpmclone-list'),
+                                    {'source_release_id': 'release-2.0',
+                                     'target_release_id': 'release-1.0'},
+                                    format='json')
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+
+    def test_clone_overridesRPM_without_source_release(self):
+        response = self.client.post(reverse('overridesrpmclone-list'),
+                                    {'target_release_id': 'release-1.0'},
+                                    format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_clone_when_source_release_has_no_overridesrpm(self):
+        args = {"name": "Supplementary", "short": "supp", "version": "1.1",
+                "release_type": "ga"}
+        target_response = self.client.post(reverse('release-list'), args)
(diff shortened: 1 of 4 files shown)


pdc/apps/compose/views.py
```diff
@@ -1276,6 +1277,110 @@ def _clear(self, release_obj, args):
...
+
+        if 'source_release_id' not in data:
+            return Response({'detail': 'Missing source_release_id'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        source_release_id = data.pop('source_release_id')
+        if 'target_release_id' not in data:
+            return Response({'detail': 'Missing target_release_id'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        target_release_id = data.pop('target_release_id')
+        try:
```
ycheng-aa: 'source_release_id' part and 'target_release_id' part are duplicate code.

pdc/apps/compose/views.py
```diff
@@ -1276,6 +1277,110 @@ def _clear(self, release_obj, args):
...
+                            status=status.HTTP_400_BAD_REQUEST)
+
+        if 'source_release_id' not in data:
+            return Response({'detail': 'Missing source_release_id'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        source_release_id = data.pop('source_release_id')
+        if 'target_release_id' not in data:
+            return Response({'detail': 'Missing target_release_id'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        target_release_id = data.pop('target_release_id')
```
ycheng-aa: 'source_release_id' part and 'target_release_id' part are duplicate code.

pdc/apps/compose/views.py
```diff
@@ -1276,6 +1277,110 @@ def _clear(self, release_obj, args):
...
+            source_release = get_object_or_404(Release, release_id=source_release_id)
+        except Http404:
+            return Response({'detail': 'Source_release %s is not existed' % source_release_id},
+                            status=status.HTTP_404_NOT_FOUND)
+        try:
+            target_release = get_object_or_404(Release, release_id=target_release_id)
+        except Http404:
+            return Response({'detail': 'Target_release %s is not existed' % target_release_id},
+                            status=status.HTTP_404_NOT_FOUND)
+
```
ycheng-aa: Above 2 are duplicate code.

pdc/apps/compose/views.py
```diff
@@ -1276,6 +1277,110 @@ def _clear(self, release_obj, args):
...
+                                                              variant=rpm.variant,
+                                                              arch=rpm.arch,
+                                                              srpm_name=rpm.srpm_name,
+                                                              comment=rpm.comment
+                                                              )
+            if created:
+                results.append(orpm.export())
+                request.changeset.add('OverridesRPM', orpm.pk,
+                                      'null', json.dumps(orpm.export()))
+            else:
```
ycheng-aa: Looks like this else is useless.

pdc/apps/compose/views.py
```diff
@@ -1276,6 +1277,110 @@ def _clear(self, release_obj, args):
...
+        for rpm in overrides_rpm:
+            orpm, created = OverrideRPM.objects.get_or_create(release=target_release,
+                                                              rpm_name=rpm.rpm_name,
+                                                              rpm_arch=rpm.rpm_arch,
+                                                              variant=rpm.variant,
+                                                              arch=rpm.arch,
+                                                              srpm_name=rpm.srpm_name,
+                                                              comment=rpm.comment
+                                                              )
+            if created:
```
ycheng-aa: Only created will be put in results, will it cause user only know which orpms are created, but don't know which orpms are belonged to target_release? I am not sure if it is a issue.

pdc/apps/compose/views.py
```diff
@@ -1276,6 +1277,110 @@ def _clear(self, release_obj, args):
...
+                results.append(orpm.export())
+                request.changeset.add('OverridesRPM', orpm.pk,
+                                      'null', json.dumps(orpm.export()))
+            else:
+                continue
+        if results:
+            return Response(status=status.HTTP_201_CREATED, data=results)
+        else:
+            return Response({'detail': 'Both releases contain the same overrides-rpm'},
+                            status=status.HTTP_400_BAD_REQUEST)
```
ycheng-aa: Is returning 400 if 2 releases have the same orpms a part of AC?

pdc/apps/compose/views.py
```diff
@@ -1276,6 +1277,110 @@ def _clear(self, release_obj, args):
...
+            if created:
+                results.append(orpm.export())
+                request.changeset.add('OverridesRPM', orpm.pk,
+                                      'null', json.dumps(orpm.export()))
+            else:
+                continue
+        if results:
+            return Response(status=status.HTTP_201_CREATED, data=results)
+        else:
+            return Response({'detail': 'Both releases contain the same overrides-rpm'},
```
simozhan: it seems if target overrides-rpm >= source overrides-rpm will go into this case. so the error message is not correct

pdc/apps/compose/tests.py
```diff
@@ -730,6 +730,98 @@ def test_delete_two_by_id(self):
...
+        args = {"name": "Supplementary", "short": "supp", "version": "1.1",
+                "release_type": "ga"}
+        target_response = self.client.post(reverse('release-list'), args)
+        self.assertEqual(target_response.status_code, status.HTTP_201_CREATED)
+        target_release_id = target_response.data['release_id']
+        response = self.client.post(reverse('overridesrpmclone-list'),
+                                    {'source_release_id': 'release-1.0',
+                                     'target_release_id': target_release_id},
+                                    format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
simozhan: there is no assertNumChanges in the test when overridesrpm created

bliuredhat: okay, I will add it.

pdc/apps/compose/views.py
bliuredhat: I consider that it's not duplicated here, because  need  the object which return from get_object_or_404()  for other code.

pdc/apps/compose/views.py
bliuredhat: Yes, you are right && good issue !  We can discuss it in meeting. In my opinion, the current result may be more reasonable. @simozhan and @lubomir; please take some time to care about this issue and share you ideal, thank you.

pdc/apps/compose/views.py
bliuredhat: yes. I think so.

pdc/apps/compose/views.py
bliuredhat: Great~!

pdc/apps/compose/views.py
bliuredhat: thank you; 200 may be okay.

pdc/apps/compose/views.py
lubomir: I think returning just the created overrides is ok. The case with cloning to a release which already has some is very rare in my opinion.

pdc/apps/compose/views.py
lubomir: I agree with @ycheng-aa that if this logic could be abstracted into a separate function it would be better.

Anyway, using `get_object_or_404` does not have any advantages here. Why not use `Release.objects.get()` directly and catch `Release.DoesNotExist`?

Grammar note: the error should be `Target release %s does not exist.`.

lubomir:
I'm not sure I understand the commit message correctly. I think it should drop the `To`. Having just ` Create rpc/overrides/rpm/clone` would be nice enough.

pdc/apps/compose/views.py
bliuredhat: @lubomir , thank you~! Yes, I have interchanged ideas with @ycheng-aa after grooming meeting yesterday; I also consider his ideal is right,  which duplicated code existed.  BTW, your suggestion may be more simple and clear.

pdc/apps/compose/tests.py
```diff
@@ -839,6 +848,29 @@ def test_clone_overridesrpm_with_error_keys(self):
...
+        target_release_id = target_response.data['results'][1]['release_id']
+
+        response2 = self.client.get(reverse('releaserpmmapping-detail',
+                                            args=[target_release_id, 'bash']))
+        self.assertEqual(response2.status_code, status.HTTP_404_NOT_FOUND)
+        response3 = self.client.post(reverse('overridesrpmclone-list'),
+                                     {'source_release_id': 'release-1.0',
+                                     'target_release_id': target_release_id},
+                                     format='json')
+        self.assertEqual(response3.status_code, status.HTTP_201_CREATED)
```
ycheng-aa: Maybe it's not very necessary use response1, response2, response3, looks like just response is fine. It seems all response will be used only once.

ycheng-aa:
Generally looks good for me.

simozhan:
Looks good to me

PR #326, opened 2016-03-24
Stop truncating arch names on compose page

bliuredhat:
JIRA:PDC-1226


lubomir:
:+1: Maybe clarify the commit message a little: `Stop truncating arch names on compose page`?

simozhan:
Yes. This commit message is better. :+1:

PR #328, opened 2016-03-29
Sorted permission list with the same api name

bliuredhat:
JIRA: PDC-1066

### DIFF
--- a/pdc/apps/auth/models.py
+++ b/pdc/apps/auth/models.py
@@ -5,7 +5,7 @@
 # http://opensource.org/licenses/MIT
 #
 
-from django.contrib.auth.models import PermissionsMixin, UserManager, AbstractBaseUser, _user_get_all_permissions
+from django.contrib.auth.models import PermissionsMixin, UserManager, AbstractBaseUser
 from django.db import models
 from django.core import validators
 from django.utils.translation import ugettext_lazy as _
@@ -84,5 +84,4 @@ class User(AbstractBaseUser, PermissionsMixin):
         send_mail(subject, message, from_email, [self.email], **kwargs)
 
     def get_all_permissions(self, obj=None):
-        tmp_permission = _user_get_all_permissions(self, obj)
-        return sorted(tmp_permission)
+        return sorted(super(User, self).get_all_permissions(obj))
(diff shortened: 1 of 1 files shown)


pdc/apps/auth/models.py
```diff
@@ -82,3 +82,7 @@ def email_user(self, subject, message, from_email=None, **kwargs):
         Sends an email to this User.
         """
         send_mail(subject, message, from_email, [self.email], **kwargs)
+
+    def get_all_permissions(self, obj=None):
+        tmp_permission = _user_get_all_permissions(self, obj)
+        return sorted(tmp_permission)
```
ycheng-aa: _user_get_all_permissions looks like it want to be a private method. Is it appropriate to use here? Is there get_all_permissions in parent class we can call it to get result and sort?

BTW: this 2 line also could be
```
return sorted(_user_get_all_permissions(self, obj))
```

simozhan: Agree with @ycheng-aa one line is better

bliuredhat: @simozhan and @ycheng-aa ; Thank you~!

bliuredhat:
@simozhan and @ycheng-aa ; please help to review this patch; thank you~!

pdc/apps/auth/models.py
```diff
@@ -84,5 +84,5 @@ def email_user(self, subject, message, from_email=None, **kwargs):
         send_mail(subject, message, from_email, [self.email], **kwargs)
 
     def get_all_permissions(self, obj=None):
-        tmp_permission = _user_get_all_permissions(self, obj)
+        tmp_permission = PermissionsMixin.get_all_permissions(self, obj)
```
ycheng-aa: Because PermissionsMixin already in User's class inheritance hierarchy, should it use `super` to access `get_all_permissions`? So if one day use other mixin, it will automatically use new mixin's `get_all_permissions`.

And again, these 2 lines may be replaced by one line.

ycheng-aa:
Looks good．

BTW: It should be `Store` instead of `Sorte` in title and commit message.

simozhan:
looks good

PR #331, opened 2016-04-05
Dsable eus/aus repo checks

bliuredhat:
When converting data from ET to PDC, the restrictions on repo names (which are there for a good reason) don't meet with historical data:
* Updates releases can contain EUS and AUS repos
* EUS releases can contain AUS repos

The checks to be modified or disabled in order to be able to import the data.

JIRA: PDC-1426

### DIFF
--- a/pdc/apps/repository/models.py
+++ b/pdc/apps/repository/models.py
@@ -6,7 +6,6 @@
 #
 
 
-from django.core.exceptions import ValidationError
 from django.db import models
 
 from pdc.apps.common.models import get_cached_id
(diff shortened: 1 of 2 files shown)


pdc/apps/repository/models.py
```diff
@@ -108,18 +108,6 @@ def clean(self):
...
-        if release_type != "eus" and ".z" in self.name:
-            raise ValidationError("Found '.z' in repo name '%s', but release type is '%s'"
-                                  % (self.name, release_type))
-
-        if release_type == "aus" and (".aus" not in self.name and ".ll" not in self.name):
-            raise ValidationError("Missing '.aus' or '.ll' in repo name '%s'" % self.name)
-        if release_type != "aus" and (".aus" in self.name or ".ll" in self.name):
-            raise ValidationError("Found '.aus' or '.ll' in repo name '%s', but release type is '%s'"
-                                  % (self.name, release_type))
-
```
sochotnicky: I believe we said these checks would be based on configuration option - not deleted but being able to turn them on/off

bliuredhat: @sochotnicky; yes, you are right~!

bliuredhat:
@sochotnicky  and @lubomir ; please help to review this patch, Thank you !

lubomir:
First, the flag name should describe what it actually configures: `ENABLE_REPO_NAME_VALIDATION` would be better.

However, I don't think that it should be configurable though. This is an internal thing after all: Fedora may have different needs. I would prefer to remove the repo name validations completely. We can introduce them again as an internal extension with some exceptions for old releases.

bliuredhat:
@lubomir and @sochotnicky ; Thank you all ! There is a conflict between your ideal,  @Stano prefer to keep the repo validation and based on configuration option, which able to turn on/off;  @lubomir prefer to remove the repo validation and as an internal extension . 
It's difficult for me to decide which way is more better, can you help me ?

bliuredhat:
@sochotnicky  and @lubomir ; please help to review this patch.

sochotnicky:
We want to keep as many things as possible in this repo. Otherwise we are just adding work for ourselves. If it's not causing issues - please don't remove the checks

I would also prefer if the config variable was called something more clear (and correct) than ET_TO_PDC_REPO_FLAGE. Maybe "RELEASE_TYPE_EUS_NAME_CHECKS" would be appropriate

bliuredhat:
@sochotnicky;  your ideal is reasonable~! Maybe  the best option is to remove that code from upstream version completely; if need,  we can introduce some configurable checks in rhpdc.  these validations are only useful internally, in Fedora this makes no sense.

simozhan:
We discussed in the meeting that this code can be removed and using other service to do the validation in the future. So this change looks good to me.

ycheng-aa:
:+1: 

It's better to rebase the pull request to 1 commit and add detailed commit message for future tracking. As discussed, the logic is gonna be implemented in somewhere else.

PR #335, opened 2016-04-06
Return warning when query in boolean field with illegal value.

ycheng-aa:
JIRA: PDC-1392


simozhan:
looks good to me.

bliuredhat:
looks good for me

PR #338, opened 2016-04-08
rebrand

nphilipp:
This is mainly so that the home page doesn't single out Red Hat.

"...which are required to support automation of {Red Hat -> Fedora}
engineering workflows..."

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ FROM fedora:21
 MAINTAINER Zhikun Lao <zlao@redhat.com>
 
 LABEL Description = "product-definition-center"
-LABEL Vendor = "Red Hat"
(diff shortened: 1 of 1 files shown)


lubomir:
I don't think changing the Dockerfile will do anything. As far as I know, that is not used anywhere.

You probably want to change [pdc/templates/home/index.html](https://github.com/product-definition-center/product-definition-center/blob/master/pdc/templates/home/index.html). Red Hat is using a customized template, so any changes should be fine.

nphilipp:
Ahh thanks—conversation with @ralphbean on IRC yesterday made me believe that this is filled from the label defined in the docker file. I'll make a different patch then.

nphilipp:
New PR is #339.

PR #340, opened 2016-04-12
Fix the bug about "Manage Overrides" on release without compose

bliuredhat:
JIRA: PDC-1421


lubomir:
This should fix it :+1:

simozhan:
Looks good to me.

ycheng-aa:
Looks good. 

Maybe add a test case  to cover it even better.

bliuredhat:
@ycheng-aa ; you are right, I add the test case~.

PR #343, opened 2016-04-14
Add doc to note the user when PUT optional parameter.

simozhan:
JIRA: PDC-1456

Only if the field in serializer is explicitly setting default value will make this happen.

### DIFF
--- a/pdc/apps/common/constants.py
+++ b/pdc/apps/common/constants.py
@@ -21,3 +21,9 @@ ARCH_SRC = 'src'
 
 # PDC warning field in response header
 PDC_WARNING_HEADER_NAME = 'pdc-warning'
+
+# Warning when using put method on optional parameter
+PUT_OPTIONAL_PARAM_WARNING = {
+    'PUT_OPTIONAL_PARAM_WARNING': 'When using the `PUT` method, if an optional field is not specified' +
+                                  ' in the input, it will be erased.',
+}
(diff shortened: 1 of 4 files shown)


bliuredhat:
looks good for me

pdc/apps/common/constants.py
```diff
@@ -21,3 +21,9 @@
 
 # PDC warning field in response header
 PDC_WARNING_HEADER_NAME = 'pdc-warning'
+
+# Warning when using put method on optional parameter
+PUT_OPTIONAL_PARAM_WARNING = {
+    'PUT_OPTIONAL_PARAM_WARNING': 'When using the `PUT` method, if an optional field is not specified' +
+                                  ' in the input, it will be erased.',
```
ycheng-aa: Do you think original text "it will be erased" is accurate? For me, describe it as something like change back to default value maybe easier for me to understand. But I am not sure, maybe original text is more precise.

ycheng-aa:
:+1:

pdc/apps/common/constants.py
simozhan: i think the original text has the meaning that "if the optional field has a default value, it will be reset to it", but, yes, maybe your description is more clear. @lubomir is that ok to change the description in this way "it will be erased" -> "it will be changed back to default value"

simozhan:
Stay with the original description, if future someone argue with it, let's change it then.

PR #348, opened 2016-04-27
Get compose path based on just compose_id/location/schema

ycheng-aa:
Add a new end point: compose-locations/(?P<compose_id>[^/]+)/
(?P<location_id>[^/]+)/(?P<schema>[^/]+) to get compose location
according to compose tree location's url.
If the url data is inconsistent, one of them will be returned
as result and pdc-warning will be returned in HTTP header.

JIRA: PDC-1346

### DIFF
--- a/pdc/apps/compose/views.py
+++ b/pdc/apps/compose/views.py
@@ -1956,3 +1956,45 @@ class ComposeTreeViewSet(ChangeSetModelMixin,
             STATUS: 204 NO CONTENT
         """
         return super(ComposeTreeViewSet, self).destroy(request, *args, **kwargs)
+
+
+class ComposeLocationViewSet(viewsets.GenericViewSet):
+    """
+    API endpoint that allows querying compose-variant-arch relevant to location.
+
+    ##Test tools##
+
+    You can use ``curl`` in terminal, with -X GET,
+    or GUI plugins for browsers, such as ``RESTClient``, ``RESTConsole``.
+    """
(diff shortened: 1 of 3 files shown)


pdc/apps/compose/views.py
```diff
@@ -1956,3 +1956,45 @@ def destroy(self, request, *args, **kwargs):
...
+        compose_trees = ComposeTree.objects.filter(compose__compose_id=kwargs['compose_id'],
+                                                   location__short=kwargs['location_id'],
+                                                   scheme__name=kwargs['schema'])
+        if not compose_trees:
+            result = Response(status=status.HTTP_404_NOT_FOUND,
+                              data={
+                                  'detail': "Can't find compose tree with compose id '%s', location '%s', schema '%s'" %
+                                            (kwargs['compose_id'], kwargs['location_id'], kwargs['schema'])})
+        else:
+            url_set = set(map(lambda obj: obj.url, compose_trees))
```
lubomir: Would a list comprehension be more readable here?
```python
set([obj.url for obj in compose_trees])
```

What do you think about returning the `Response` directly instead of using the `result` variable`?

ycheng-aa: Change to comprehension in 3db97198940d4f17. It's more readable. Thanks.

If return Response directly, it will have 3 exits. For me maybe use `result` to keep it one exit is also acceptable.

bliuredhat:
looks good for me~!

PR #350, opened 2016-05-02
Enable query with multi values.

simozhan:
It doesn't make sense to enable multi value for following param.
1. Boolean param.
2. Param with substring match.
3. Param with few kinds of type.

NullableCharFilter and MethodFilter with special filter logic
are not enabled as well.

JIRA: PDC-1459


ycheng-aa:
:+1:

bliuredhat:
:+1:

PR #356, opened 2016-05-10
Add two new rules to API stability

sochotnicky:
- default values of optional arguments

  If default value of optional argument changes, existing API users could
  suddenly supply incomplete/incorrect information for their use case

- order of returned results

  If order of multiple results changes it can affect how it's interpreted by the
  clients. For example releases returned for given product ID, or composes for
  given release ID.

Change-Id: Icfc10ce5db2d90fae56e0debbcc2476a11494471

### DIFF
--- a/docs/source/using_api.rst
+++ b/docs/source/using_api.rst
@@ -23,6 +23,11 @@ In this context, stable means:
  * Arguments for APIs will not be removed or renamed
  * Keys in returned JSON dictionaries will not be removed or renamed
  * If new features are added to these APIs – which is quite possible – they will not break or change the meaning of existing methods. In other words, “stable” does not (necessarily) mean *complete*
+ * Default values of optional arguments will not change
+ * Order of returned results will not change for following results:
+    * releases in releases resource API
+    * composes inside compose_set in releases resource API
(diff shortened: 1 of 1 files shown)


docs/source/using_api.rst
```diff
@@ -23,6 +23,8 @@ In this context, stable means:
  * Arguments for APIs will not be removed or renamed
  * Keys in returned JSON dictionaries will not be removed or renamed
  * If new features are added to these APIs – which is quite possible – they will not break or change the meaning of existing methods. In other words, “stable” does not (necessarily) mean *complete*
+ * Default values of optional arguments will not change
+ * Order of returned results will not change
```
lubomir: I assume this only applies to place where the order is documented – lists of releases or composes. I don't think we want to guarantee the order in places that are not actually ordered.

sochotnicky: I guess it would make sense to make this more explicit. I will expand this part

ycheng-aa:
:+1:

lubomir:
:+1:  By the way, you can use \`\`code\`\` to mark things.

PR #358, opened 2016-05-16
Create release-group API

bliuredhat:
In errata tool, it want to group product versions
(called releases in PDC) into "release groups"; which need
an equivalent of this functionality to properly migrate RCM off
of Errata configuration.

JIRA: PDC-1451

### DIFF
--- a/pdc/apps/release/models.py
+++ b/pdc/apps/release/models.py
@@ -372,3 +372,39 @@ class VariantArch(models.Model):
 
     def __unicode__(self):
         return u"%s.%s" % (self.variant, self.arch)
+
+
+class ReleaseGroupType(models.Model):
+    name = models.CharField(max_length=50, unique=True)
+
+    def __unicode__(self):
+        return u"%s" % self.name
+
+
+class ReleaseGroup(models.Model):
+    name                = models.CharField(max_length=255, blank=False, unique=True)
+    description         = models.CharField(max_length=255, blank=False)
+    type                = models.ForeignKey(ReleaseGroupType)
+    releases            = models.ManyToManyField(Release, blank=True)
+    active              = models.BooleanField(default=True)
+
+    class Meta:
+        ordering = ("name", )
+
+    def __unicode__(self):
+        return u"%s" % self.name
+
+    def export(self, fields=None):
+        _fields = ['name', 'description', 'type', 'releases', 'active'] if fields is None else fields
+        result = dict()
+        for field in ('name', 'description', 'active'):
+            if field in _fields:
+                result[field] = getattr(self, field)
+        if 'releases' in _fields:
+            result['releases'] = []
+            releases = self.releases.all()
+            for obj in releases:
+                    result['releases'].append(obj.export())
+        if 'type' in _fields:
+            result['type'] = self.type.name
+        return result
--- a/pdc/apps/release/routers.py
+++ b/pdc/apps/release/routers.py
@@ -14,6 +14,7 @@ router.register(r'product-versions', views.ProductVersionViewSet)
 router.register(r'releases', views.ReleaseViewSet)
 router.register(r'base-products', views.BaseProductViewSet)
 router.register(r'release-types', views.ReleaseTypeViewSet, base_name='releasetype')
+router.register(r'release-groups', views.ReleaseGroupsViewSet, base_name='releasegroups')
 router.register('releases/(?P<release_id>[^/]+)/rpm-mapping',
                 views.ReleaseRPMMappingView,
                 base_name='releaserpmmapping')
--- a/pdc/apps/release/serializers.py
+++ b/pdc/apps/release/serializers.py
@@ -9,7 +9,9 @@ from django.core.exceptions import FieldError
 from pdc.apps.common.fields import ChoiceSlugField
 from pdc.apps.common import models as common_models
 from pdc.apps.common.serializers import StrictSerializerMixin
-from .models import Product, ProductVersion, Release, BaseProduct, ReleaseType, Variant, VariantArch, VariantType
+from .models import (Product, ProductVersion, Release,
+                     BaseProduct, ReleaseType, Variant,
+                     VariantArch, VariantType, ReleaseGroup, ReleaseGroupType)
 from . import signals
 
 
@@ -205,3 +207,27 @@ class VariantTypeSerializer(StrictSerializerMixin, serializers.ModelSerializer):
     class Meta:
         model = VariantType
         fields = ('name',)
+
+
+class ReleaseGroupSerializer(StrictSerializerMixin, serializers.ModelSerializer):
+    description         = serializers.CharField(required=True)
+    type                = serializers.SlugRelatedField(slug_field='name',
+                                                       queryset=ReleaseGroupType.objects.all())
+    releases            = ChoiceSlugField(slug_field='release_id',
+                                          many=True, queryset=Release.objects.all(),
+                                          allow_null=True, required=False)
+    active              = serializers.BooleanField(default=True)
+
+    class Meta:
+        model = ReleaseGroup
+        fields = ('name', 'description', 'type', 'releases', 'active')
+
+    def to_internal_value(self, data):
+        releases = data.get('releases', [])
+        for release in releases:
+            try:
+                Release.objects.get(release_id=release)
+            except Release.DoesNotExist:
+                raise serializers.ValidationError({'detail': 'release %s does not exist' % release})
+
+        return super(ReleaseGroupSerializer, self).to_internal_value(data)
--- a/pdc/apps/release/views.py
+++ b/pdc/apps/release/views.py
@@ -20,7 +20,7 @@ from .forms import (ReleaseSearchForm, BaseProductSearchForm,
 from .serializers import (ProductSerializer, ProductVersionSerializer,
                           ReleaseSerializer, BaseProductSerializer,
                           ReleaseTypeSerializer, ReleaseVariantSerializer,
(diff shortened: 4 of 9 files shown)


pdc/apps/release/models.py
```diff
@@ -372,3 +372,42 @@ class Meta:
...
+
+    def export(self, fields=None):
+        _fields = ['name', 'description', 'type', 'releases'] if fields is None else fields
+        result = dict()
+        if 'name' in _fields:
+            result['name'] = self.name
+        if 'description' in _fields:
+            result['description'] = self.description
+        if 'active' in _fields:
+            result['active'] = self.active
```
ycheng-aa: Above 3 `if` looks duplicate. Could it be something like:
```
for i in ('name', 'description', 'active'):
    if i in _fields:
        result[i] = getattr(self, i)
```

pdc/apps/release/routers.py
```diff
@@ -14,6 +14,8 @@
 router.register(r'releases', views.ReleaseViewSet)
 router.register(r'base-products', views.BaseProductViewSet)
 router.register(r'release-types', views.ReleaseTypeViewSet, base_name='releasetype')
+router.register(r'release-group-types', views.ReleaseGroupTypeViewSet, base_name='releasegrouptypes')
```
ycheng-aa: Should PDC provide the api for release-group-types? Maybe I remember wrong. But In my impression, @sochotnicky mentioned no need to provide API for types.

pdc/apps/release/views.py
```diff
@@ -1013,3 +1014,134 @@ def list(self, request, *args, **kwargs):
...
+        %(SERIALIZER)s
+        """
+        return super(ReleaseGroupTypeViewSet, self).list(*args, **kwargs)
+
+
+class ReleaseGroupsViewSet(ChangeSetModelMixin,
+                           StrictQueryParamMixin,
+                           viewsets.GenericViewSet):
+    """
+    API endpoint that allows release_group_types to be viewed or edited.
```
ycheng-aa: Not `release_group_types`.

pdc/apps/release/models.py
```diff
@@ -372,3 +372,42 @@ class Meta:
...
+    active              = models.BooleanField(default=True)
+
+    class Meta:
+        ordering = ("name", )
+
+    def __unicode__(self):
+        return u"%s" % self.name
+
+    def export(self, fields=None):
+        _fields = ['name', 'description', 'type', 'releases'] if fields is None else fields
```
simozhan: seems no 'active' here, typo?

pdc/apps/release/routers.py
simozhan: Agree with @ycheng-aa, @sochotnicky mentioned it in the ticket's comment.

bliuredhat: @ycheng-aa ; your ideal is right, and I hope we can discuss this issue in meeting.  In my opinion, when we create a instance of release-group, need to verify the Type para would have been existed in DB, So this may be useful for user. If we don't it, it would be removed.

pdc/apps/release/views.py
```diff
@@ -1013,3 +1014,134 @@ def list(self, request, *args, **kwargs):
...
+            return Response({'detail': '%s keys are not allowed' % list(extra_keys)},
+                            status=status.HTTP_400_BAD_REQUEST)
+        if 'releases' in data:
+            for key in data['releases']:
+                try:
+                    models.Release.objects.get(release_id=key)
+                except models.Release.DoesNotExist:
+                    return Response({'detail': '%s release does not exist' % key},
+                                    status=status.HTTP_404_NOT_FOUND)
+
```
ycheng-aa: I think you already defined this logic in serialize. And I think this logic should be in serializer.

pdc/apps/release/views.py
```diff
@@ -1013,3 +1014,134 @@ def list(self, request, *args, **kwargs):
...
+        if 'type' in data:
+            try:
+                models.ReleaseGroupType.objects.get(name=data['type'])
+            except models.ReleaseGroupType.DoesNotExist:
+                return Response({'detail': '%s type does not exist' % data['type']},
+                                status=status.HTTP_404_NOT_FOUND)
+
```
ycheng-aa: Same as above.

pdc/apps/release/views.py
```diff
@@ -1013,3 +1014,134 @@ def list(self, request, *args, **kwargs):
...
+        %(SERIALIZER)s
+        """
+        data = request.data
+        keys = set(['description', 'name', 'type'])
+        arg_filter_map = ['active', 'releases']
+        allowed_keys = list(keys) + arg_filter_map
+        extra_keys = set(data.keys()) - set(allowed_keys)
+        if extra_keys:
+            return Response({'detail': '%s keys are not allowed' % list(extra_keys)},
+                            status=status.HTTP_400_BAD_REQUEST)
```
ycheng-aa: After `fields` defined in serializer, do we need this? What is `keys` and `arg_filter_map` stand for here?

pdc/apps/release/views.py
bliuredhat: @ycheng-aa ; yes, which can verify the logic in serializer, but the error info is not usable, I think QE prefer to get the simple and clear error info from API.

pdc/apps/release/views.py
bliuredhat: thank you,  `keys ` and `arg_filter_map` is not need here, I will remove it.
BTW,  I consider these logic checking is prepare for QE, which can return a clear info as we want.

pdc/apps/release/views.py
ycheng-aa: Maybe you can customize error message in serializer, or define you own field in serializer.  I saw a lot of API use serializer ChoiceSlugField or SlugRelatedField directly. I am not quite sure why their error info is not usable here. Anyway, the behavior in serializer could be customized.

pdc/apps/release/serializers.py
```diff
@@ -205,3 +207,34 @@ class VariantTypeSerializer(StrictSerializerMixin, serializers.ModelSerializer):
...
+    def to_internal_value(self, data):
+        releases = data.get('releases', None)
+        if releases:
+            for release in releases:
+                try:
+                    Release.objects.get(release_id=release)
+                except Release.DoesNotExist:
+                    raise serializers.ValidationError({'detail': '%s release does not exist' % release})
+
+        type = data.get('type', None)
```
lubomir: `type` is a builtin function. Maybe using a different name would avoid some confusion. For the class attribute (on model or serializer) it's probably fine, as that will almost always be used with an instance.

pdc/apps/release/serializers.py
```diff
@@ -205,3 +207,34 @@ class VariantTypeSerializer(StrictSerializerMixin, serializers.ModelSerializer):
...
+                                          many=True, queryset=Release.objects.all(),
+                                          allow_null=True, required=False)
+    active              = serializers.BooleanField(default=True)
+
+    class Meta:
+        model = ReleaseGroup
+        fields = ('name', 'description', 'type', 'releases', 'active')
+
+    def to_internal_value(self, data):
+        releases = data.get('releases', None)
```
lubomir: If the fallback was `[]` instead of `None`, the `if` condition would no longer be necessary.

pdc/apps/release/views.py
lubomir: The docstring should mention here the endpoint is experimental.

pdc/apps/release/serializers.py
bliuredhat: @lubomir ; thank you; I will update it .

pdc/apps/release/views.py
bliuredhat: @lubomir ; I have removed the release_group_types here.

pdc/apps/release/serializers.py
```diff
@@ -205,3 +207,33 @@ class VariantTypeSerializer(StrictSerializerMixin, serializers.ModelSerializer):
...
+                Release.objects.get(release_id=release)
+            except Release.DoesNotExist:
+                raise serializers.ValidationError({'detail': '%s release does not exist' % release})
+
+        type_key = data.get('type', None)
+        if type_key:
+            try:
+                ReleaseGroupType.objects.get(name=type_key)
+            except ReleaseGroupType.DoesNotExist:
+                raise serializers.ValidationError({'detail': '%s type does not exist' % type_key})
```
ycheng-aa: I think type you should use it's default behavior. It's better to show the legal value. Maybe you could implement `to_internal_value` in field level. Here maybe release field.

pdc/apps/release/serializers.py
```diff
@@ -205,3 +207,27 @@ class VariantTypeSerializer(StrictSerializerMixin, serializers.ModelSerializer):
...
+        model = ReleaseGroup
+        fields = ('name', 'description', 'type', 'releases', 'active')
+
+    def create(self, validated_data):
+        releases = validated_data.pop('releases', [])
+        for release in releases:
+            try:
+                Release.objects.get(release_id=release)
+            except Release.DoesNotExist:
+                raise serializers.ValidationError({'detail': '%s release does not exist' % release})
```
simozhan: Is it better "Release %s does not exist"

pdc/apps/release/views.py
simozhan: i guess @lubomir  means add docstring in ReleaseGroupsViewSet

ycheng-aa:
Looks good for me.

simozhan:
looks good to me

PR #359, opened 2016-05-16
Implement resource based permissions control.

ycheng-aa:
1. Provide 3 APIs, resource-permissions,
   current-user-resource-permissions and
   group-resource-permissions.
2. Resources are APIs which need to be included in permission
   control. Permissions are creat, read, update, delete. PDC
   will collect resource permissions accroding to router
   automatically.
3. resource-permissions API provide list function to
   list all available resource permissions.
   current-user-resource-permissions list current
   user's resource permissions.
4. Through group-resource-permissions, resource permission
   could be assigned to or removed from a group. Users in
   that group will be affected.
5. Prevent non-superuser from accessing admin page.

JIRA: PDC-1384

### DIFF
--- a/pdc/apps/auth/migrations/0004_auto_20160512_1051.py
+++ b/pdc/apps/auth/migrations/0004_auto_20160512_1051.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def create_action(apps, schema_editor):
+    action_permission = apps.get_model('kerb_auth', 'ActionPermission')
+    for action in ('create', 'read', 'update', 'delete'):
+        action_permission.objects.create(name=action)
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('kerb_auth', '0003_auto_20151126_0811'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='ActionPermission',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('name', models.CharField(unique=True, max_length=100)),
+            ],
+        ),
+        migrations.CreateModel(
+            name='GroupResourcePermission',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('group', models.ForeignKey(to='auth.Group')),
+            ],
+        ),
+        migrations.CreateModel(
+            name='Resource',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('name', models.CharField(unique=True, max_length=500)),
+                ('view', models.CharField(max_length=1000)),
+            ],
+        ),
+        migrations.CreateModel(
+            name='ResourcePermission',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('permission', models.ForeignKey(to='kerb_auth.ActionPermission')),
+                ('resource', models.ForeignKey(to='kerb_auth.Resource')),
+            ],
+        ),
+        migrations.AddField(
+            model_name='groupresourcepermission',
+            name='resource_permission',
+            field=models.ForeignKey(to='kerb_auth.ResourcePermission'),
+        ),
+        migrations.RunPython(create_action)
+    ]
--- a/pdc/apps/auth/views.py
+++ b/pdc/apps/auth/views.py
@@ -24,7 +24,10 @@ from rest_framework import mixins, viewsets
 from . import backends
 from . import filters
 from . import serializers
+from . import models
+from pdc.apps.auth.permissions import APIPermission
 from pdc.apps.common.viewsets import StrictQueryParamMixin, ChangeSetUpdateModelMixin
+from pdc.apps.common import viewsets as common_viewsets
 from pdc.apps.utils.utils import group_obj_export
 
 
@@ -483,6 +486,7 @@ class GroupViewSet(ChangeSetUpdateModelMixin,
     queryset = Group.objects.all().order_by('id')
     serializer_class = serializers.GroupSerializer
     filter_class = filters.GroupFilter
+    permission_classes = (APIPermission,)
     Group.export = group_obj_export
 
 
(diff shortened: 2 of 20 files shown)


pdc/apps/auth/migrations/0004_auto_20160512_1051.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def create_action(apps, schema_editor):
+    """Migrate contacts stored directly on release components."""
```
lubomir: This comment does not seem to match the method.

pdc/apps/auth/views.py
```diff
@@ -527,3 +530,134 @@ def list(self, request):
...
+        resource_permission_set = set([])
+        for group_resource_permission in queryset:
+            resource_permission_set.add(group_resource_permission.resource_permission)
+        serializer = serializers.ResourcePermissionSerializer(resource_permission_set, many=True)
+        return Response(serializer.data)
+
+
+class GroupResourcePermissionViewSet(common_viewsets.PDCModelViewSet):
+    """
+    This end-point provides t group resource permissions.
```
lubomir: Stray *t*.

pdc/apps/auth/views.py
```diff
@@ -527,3 +530,134 @@ def list(self, request):
...
+        __URL__: $LINK:groupresourcepermissions-list$
+
+        __Query params__:
+
+        %(FILTERS)s
+
+
+        __Response__:
+
+
```
lubomir: Is response description missing here?

pdc/apps/utils/middleware.py
```diff
@@ -27,3 +34,50 @@ def process_response(self, request, response):
...
+
+
+class RestrictAdminMiddleware(object):
+    """
+    Restricts access to the admin page to only logged-in users with a certain user-level.
+    """
+    def process_request(self, request):
+        if request.user.is_authenticated() and request.path.startswith(reverse('admin:index')):
+            if not (request.user.is_active and request.user.is_superuser):
+                return HttpResponseRedirect("/%s%s/" % (settings.REST_API_URL, settings.REST_API_VERSION))
```
lubomir: Why is this needed? The behaviour up to now was to allow login to admin section for users who are marked as *staff*.

pdc/apps/auth/views.py
```diff
@@ -527,3 +530,134 @@ def list(self, request):
...
+    def list(self, request):
+        """
+        Get information about current user.
+
+        __Method__: `GET`
+
+        __URL__: $LINK:currentuserpermissions-list$
+
+        __Response__:
+            [
```
lubomir: This does not render correctly. Maybe a blank line is missing?

lubomir:
The code looks reasonable to me. I'm not a big fan of adding another dependency, but if it makes things easier, then why not.

I think it would be nice to have the new urls grouped under the `auth/` prefix. How about removing the `resource` keyword from the URL too? Does it add any value?

A couple notes from my testing:
 * Running a development server with this patch suddenly requires login to access any API. How can the server be configured to not require any authentication for read access? This is a requirement for deploying in Fedora infrastructure.
 * Even though I'm an admin, visiting `/current-user-resource-permission` does not show anything. Furthermore, why add another end-point here instead of merging this into existing `/auth/current-user`?
 * Urls in documentation for new APIs don't show correctly.
 * Group resource permissions endpoint does not have *Create* tab, but it has *Bulk Create*.

pdc/apps/utils/middleware.py
ycheng-aa: I thought if we grant all model permissions to all groups, then API permissions could work all by itself. PDC will not have the situation that it grant a group a API's permission, but still can't work because lack of model permission. But it will bring the risk that user could manipulate model directly to avoid permission control.

I think you mentioned we should remove model permissions. But I am not sure what it means.  Did you mean remove `django.contrib.auth.backends.ModelBackend` from `AUTHENTICATION_BACKENDS`? I tested to remove it and it looks we can't login admin page.

pdc/apps/auth/migrations/0004_auto_20160512_1051.py
ycheng-aa: Fixed in 71b3211d41ee6bd4

pdc/apps/auth/views.py
ycheng-aa: Fixed in 71b3211d41ee6bd4

pdc/apps/auth/views.py
ycheng-aa: Fixed in 71b3211d41ee6bd4

pdc/apps/auth/views.py
ycheng-aa: Fixed in 71b3211d41ee6bd4

ycheng-aa:
@lubomir  Thanks you very much for the careful review.

* I moved new endpoints under `auth/`. Merged resource permission content to `/auth/current-user`.
* For superuser, It will show all resource permissions even not explicitly assigned to him and doesn't   exist in DB.
* Fixed documentation and make sure `create` tab work.
* In settings.py, add a flag to allow anyone has read permission.

lubomir:
Regarding the model permissions, I think it does not matter if they are used or not as long as they are not exposed to users (including admins). It seems confusing to have to different permission systems. What is the expected workflow?

ycheng-aa:
@lubomir I am hoping we could grant all model permissions to all groups. Then use resource based permissions control only.  This is my expected workflow.
Because user could have all model permissions, grant any resource permission to a user could make sure user really has the permission. Normal user also couldn't modify model directly instead of by accessing API. 
All this just my conjecture, I am not sure if there is something wrong in the workflow.

lubomir:
That sounds reasonable to me. If everyone has all the model permissions, are changes in the admin section still limited to admins only?

lubomir:
Anyway, my proposal here would be to merge this pull request and deploy it on devel server. Then we can test how it behaves, talk to users and have them try it. We have quite some time until next release to fix any issues.

ycheng-aa:
Changes you mean the page like https://pdc.host.dev.eng.pek2.redhat.com/changes/ ? If it's this page, it's not in admin section.
I am also not sure if the admin restriction is necessary. As you said, we could deploy then test and fix problems. Thanks.

simozhan:
Looks good to me. :+1: Agree with @lubomir, merge it first and test it.

PR #365, opened 2016-05-19
Fix a bug and add a flag for resource permissions

ycheng-aa:
If this flag is set to true then all users will
disable resource permission check. E.g. disable
it in local environment test.

JIRA: PDC-1380


simozhan:
There is an issue when creating ResourcePermission. It seems no explicit unique for this model, so it is possible to add duplicate records, that will make error happen in other place. GroupResourcePermission should be the same. 
By the way, there is a flake8 error.

bliuredhat:
:+1:

simozhan:
looks good to me

PR #367, opened 2016-05-19
Announce big changes in PDC database

ycheng-aa:
If the number of a changeset's change reach
or greater threshold (CHANGESET_SIZE_ANNOUNCE)
and CHANGESET_ANNOUNCE_EMAIL is set in settings.
Then send a email with necessary content.

JIRA: PDC-1441


lubomir:
Looks good. The original intent was for `CHANGESET_ANNOUNCE_EMAIL` to actually be an e-mail address where to send the message, but your implementation using the already configured address is better. The option could even be dropped. To turn the feature off, it would be enough to not set the size.

ycheng-aa:
Fixed in 27ad21c372cbe56fe. Also fixed the test cases. Thanks.

simozhan:
:+1:

bliuredhat:
looks good for me

PR #370, opened 2016-05-19
Create UI to display resource permissions in auth profile.

simozhan:
JIRA: PDC-1381

### DIFF
--- a/pdc/apps/auth/views.py
+++ b/pdc/apps/auth/views.py
@@ -4,6 +4,7 @@
 # Licensed under The MIT License (MIT)
 # http://opensource.org/licenses/MIT
 #
+import inspect
 import json
 
 from django.shortcuts import render, redirect
@@ -28,7 +29,8 @@ from . import models
 from pdc.apps.auth.permissions import APIPermission
 from pdc.apps.common.viewsets import StrictQueryParamMixin, ChangeSetUpdateModelMixin
 from pdc.apps.common import viewsets as common_viewsets
-from pdc.apps.utils.utils import group_obj_export
+from pdc.apps.utils.SortedRouter import router
+from pdc.apps.utils.utils import group_obj_export, convert_method_to_action
 
 
 def remoteuserlogin(request):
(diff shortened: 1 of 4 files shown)


pdc/apps/auth/views.py
```diff
@@ -73,8 +73,41 @@ def logout(request):
...
+    dict = {}
+    result = []
+
+    if user.is_superuser:
+        resource_permission_set = set([obj for obj in models.ResourcePermission.objects.all()])
+    else:
+        group_id_list = [group.id for group in user.groups.all()]
+        queryset = models.GroupResourcePermission.objects.filter(group__id__in=group_id_list)
+        for group_resource_permission in queryset:
+            resource_permission_set.add(group_resource_permission.resource_permission)
```
ycheng-aa: Above part from `if`, looks almost seem as the logic in `current-user` API. Could we extract the logic to a method and reuse it in both places to eliminate duplicate code. Refactor our code a little.

ycheng-aa:
Generally looks good. It would be nice if refactor it a little to remove duplicate.

bliuredhat:
looks good for me

pdc/apps/auth/views.py
simozhan: Thanks. Fixed.

pdc/apps/auth/views.py
```diff
@@ -73,8 +73,41 @@ def logout(request):
     return HttpResponseRedirect(redirect_to)
 
 
+def _get_resource_permissions_matrix(user):
+    http_method = ['create', 'read', 'update', 'delete']
+    resource_permission_set = set([])
+    dict = {}
```
lubomir: I would like to see this variable renamed to not shadow a Python builtin. It will stop confusing syntax highlighters.

lubomir:
Generally it looks good to me. One suggestion I have (and I don't know how difficult it would be to implement) is to show different text in cells where no call is possible. For example in *composes* + *create* everyone will see *no*, but there is no such call. Maybe having an empty cell or `N/A` or something like that would make it easier to understand.

simozhan:
@lubomir yes. this is quite reasonable. fixed.

bliuredhat:
:+1:

pdc/apps/auth/views.py
```diff
@@ -73,8 +75,58 @@ def logout(request):
...
+        try:
+            index = http_method.index(v.lower())
+            row[index + 1] = value
+        except ValueError:
+            raise Exception("The action %s is not valid" % v)
+    return row
+
+
+def _get_resource_permissions_matrix(user):
+    d = {}
```
ycheng-aa: Could name `d` a more meaningful name? It's like a dict map resource name to the resource's permission list. A meaningful name looks make code easier to understand.

ycheng-aa:
Looks good.

pdc/apps/auth/views.py
simozhan: OK. fixed

bliuredhat:
:+1:

ycheng-aa:
:+1:

lubomir:
I like it, no more comments from me.

PR #371, opened 2016-05-20
Allow using mod_auth_oidc for remote user auth.

ralphbean:
We'd like to try this out in Fedora.

### DIFF
--- a/pdc/apps/auth/backends.py
+++ b/pdc/apps/auth/backends.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2015 Red Hat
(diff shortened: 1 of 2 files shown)


puiterwijk:
Thank you very much for this patch.
This looks to be in line with what Fedora Infra is now deploying.

ralphbean:
Following up here since it's been almost a year.  I know the dev team is busy with other priorities, but we would appreciate this one being looked at.

simozhan:
I don't have environment to try it. but i think the code looks quite good to me.

ycheng-aa:
Looks good.

ralphbean:
Thank you!

PR #373, opened 2016-05-23
Change the way to generate resource permissions automatically.

ycheng-aa:
Now when a special parameter 'create' set to true in
resource permissions list method. It will generate
resource permissions.
The old middleware way looks not work well in production
environment.

JIRA: PDC-1380

### DIFF
--- a/pdc/apps/auth/views.py
+++ b/pdc/apps/auth/views.py
@@ -26,6 +26,7 @@ from . import backends
 from . import filters
 from . import serializers
 from . import models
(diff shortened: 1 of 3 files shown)


bliuredhat:
looks good for me

simozhan:
Looks ok for me. but maybe we need to investigate why middleware way doesn't work.

ycheng-aa:
Thanks @simozhan . I will investigate it. Or if there is other better solution, we also could take it into consideration.

PR #376, opened 2016-05-30
Put permission control on more resources and show correct permissions.

ycheng-aa:
1. Put permission control on some resources that only
   have read method. Because PDC currently grant users
   read permission in default, these APIs should be treated
   as others.
2. If user is granted read permission by configurations,
   display the correct permissions in real.
3. auth/current-user and auth/current-user are still not
   included in permission control

JIRA: PDC-1543


simozhan:
A typo in commit message: auth/current-user and auth/current-user(auth/token)

simozhan:
Looks good to me

bliuredhat:
looks good for me

ycheng-aa:
Thanks. Fixed the typo in commit message.

PR #378, opened 2016-06-06
Create new API of compose-tree-rtt-tests

bliuredhat:
In Compose API, the paramenter of rtt_tested_architectures is deeply nested
and hard to use/filter etc. So we want to create compose-tree-rtt-tests
endpoint which would only do RTT tests - nothing else.

JIRA: PDC-1316


simozhan:
Looks good to me

ycheng-aa:
In commit message, maybe para -> parameter, noting -> nothing. I am not sure if 'only do RTT tests' is correct in English. 

Other looks good to me.

bliuredhat:
@ycheng-aa , this API is aimed to update the rtt_tested_architectures paramenter, which in Compose API.

PR #379, opened 2016-06-06
add variant-type 'module'

KarstenHopp:
required to import composes done by pungi-modularity,
https://fedoraproject.org/wiki/Modularity/Development/Composing_Modules_with_Pungi

### DIFF
--- a/pdc/apps/release/migrations/0002_auto_20150512_0719.py
+++ b/pdc/apps/release/migrations/0002_auto_20150512_0719.py
@@ -53,6 +53,7 @@ def create_variant_types(apps, schema_editor):
         'optional',
         'addon',
         'layered-product',
+        'module',
     ]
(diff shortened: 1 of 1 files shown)


lubomir:
Adding the value in the migration will only do something when new database is created, which may or may not be what you want.

I'm going to take a guess here: you are using a patched version or productmd that defines `module` as a valid variant type ([most likely here](https://github.com/release-engineering/productmd/blob/master/productmd/composeinfo.py#L79-L84)). If that is the case, I would much rather see some logic to consume the types from productmd and create them automatically. This is already done for image types and formats ([here](https://github.com/product-definition-center/product-definition-center/blob/master/pdc/apps/package/models.py#L356-L379)).

KarstenHopp:
Your guess was correct, this got fixed in productmd

PR #382, opened 2016-06-15
Add unique to Compose PathType's name field.

simozhan:
JIRA: PDC-1557

### DIFF
--- a/pdc/apps/compose/migrations/0011_remove_duplicate_pathtypes.py
+++ b/pdc/apps/compose/migrations/0011_remove_duplicate_pathtypes.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from django.db import migrations, models
+
+def remove_duplicate_pathtype(apps, schema_editor):
+    PathType = apps.get_model('compose', 'PathType')
+    ComposeRelPath = apps.get_model('compose', 'ComposeRelPath')
+    qs = PathType.objects.values('name')
+    uniq = set([])
+    dup = set([])
+    for x in qs:
+        if x['name'] not in uniq:
+            uniq.add(x['name'])
+        else:
+            dup.add(x['name'])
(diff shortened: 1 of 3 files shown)


bliuredhat:
looks good for me

pdc/apps/compose/migrations/0011_remove_duplicate_pathtypes.py
```diff
@@ -0,0 +1,35 @@
...
+    ComposeRelPath = apps.get_model('compose', 'ComposeRelPath')
+    qs = PathType.objects.values('name')
+    uniq = []
+    dup = []
+    for x in qs:
+        if x not in uniq:
+            uniq.append(x)
+        else:
+            dup.append(x)
+    dup = [dict(t) for t in set([tuple(d.items()) for d in dup])]
```
ycheng-aa: Sees a little complicate. Looks only use `name ` is enough?
Maybe some thing like:
```
unqi = set([])
dup = set([])
for x in qs:
     if x['name'] not in uniq:
          uniq.add(x['name'])
    else:
          dup.add(x['name'])
    for name in dup:
         ........................................
```

pdc/apps/compose/migrations/0012_auto_20160615_0611.py
```diff
@@ -0,0 +1,19 @@
...
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('compose', '0011_remove_duplicate_pathtypes'),
+    ]
+
```
ycheng-aa: Could combine 0011 into this migration? Is it because re-index and change content can't in one transaction so can't work?

simozhan: Thanks for the careful review. Fixed.

ycheng-aa:
:+1:

PR #383, opened 2016-06-17
Group rpm query parameters in and operation.

simozhan:
JIRA: PDC-1529

### DIFF
--- a/pdc/apps/compose/views.py
+++ b/pdc/apps/compose/views.py
@@ -443,6 +443,7 @@ class ComposeViewSet(StrictQueryParamMixin,
     queryset = Compose.objects.all().order_by('id')
     serializer_class = ComposeSerializer
     filter_class = ComposeFilter
+    filter_fields = ('srpm_name', 'rpm_name', 'rpm_arch', 'rpm_version', 'rpm_release')
     permission_classes = (APIPermission,)
     lookup_field = 'compose_id'
     lookup_value_regex = '[^/]+'
@@ -461,10 +462,26 @@ class ComposeViewSet(StrictQueryParamMixin,
         possible to sort unconditionally as get_object() will at some point
         call this method and fail unless it receives a QuerySet instance.)
         """
-        qs = super(ComposeViewSet, self).filter_queryset(qs)
+        qs = super(ComposeViewSet, self).filter_queryset(self._filter_nvras(qs))
         if getattr(self, 'order_queryset', False):
             return sorted(qs)
         return qs
+
+    def _filter_nvras(self, qs):
+        q = Q()
+        query_params = self.request.query_params
+        query_param_rpm_key_mapping = [('rpm_name', 'name'),
+                                       ('srpm_name', 'srpm_name'),
+                                       ('rpm_version', 'version'),
+                                       ('rpm_release', 'release'),
+                                       ('rpm_arch', 'arch')]
+        for query_param, rpm_key in query_param_rpm_key_mapping:
+            rpm_value = query_params.get(query_param, None)
+            s = 'variant__variantarch__composerpm__rpm__' + rpm_key
+            if rpm_value:
+                q &= Q(**{s + '__iexact': rpm_value})
+
+        return qs.filter(q).distinct()
 
     def _fill_in_cache(self, result_queryset):
         """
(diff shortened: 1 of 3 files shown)


pdc/apps/compose/views.py
```diff
@@ -461,11 +463,43 @@ def filter_queryset(self, qs):
...
+            return qs, None
+        if len(rpm_value) == 1:
+            q &= Q(**{s + '__iexact': rpm_value[0]})
+
+        return qs, q
+
+    def _filter_nvras(self, qs):
+        q = Q()
+        query_params = dict(self.request.query_params.iterlists())
+        query_parma_rpm_key_mapping = [('rpm_name', 'name'),
```
lubomir: Typo here: `parma`.

pdc/apps/compose/views.py
```diff
@@ -461,11 +463,43 @@ def filter_queryset(self, qs):
...
+        if rpm_value is None:
+            return qs, q
+        if len(rpm_value) > 1:
+            condition_list = []
+            for val in rpm_value:
+                condition_list.append(Q(**{s + '__iexact': val}))
+            qs = qs.filter(reduce(operator.or_, condition_list)).distinct()
+            return qs, None
+        if len(rpm_value) == 1:
+            q &= Q(**{s + '__iexact': rpm_value[0]})
```
lubomir: I don't understand why the case with a single value is handled differently. My current understanding of this code is this: if you specify a filter multiple times, it will be used and all other filters will be silently ignored. If each filter is specified at most once, all of them will be used. This seems quite complicated.

If multiple different filters are used, is the condition required to be satisfied by a single RPM?

simozhan:
@lubomir Thanks for the careful review. My original idea is:  first all the rpm_* related parameters can be grouped to filter with single value.  and second for the specific parameter, for example rpm_name, maybe the user will still like to query with multi value, but i don't want to group multi value for filter, maybe that's too complicated and time consuming. so i make a priority, maybe the end user is more care about rpm_name, then srpm_name...

pdc/apps/compose/views.py
ycheng-aa: From only read the documentation below, I also thoughtit should work that way @lubomir described. Except I thought the priority will take effect if multiple fields have multiple filter value. It looks straightforward and accord to intuition.

The current implementation is filter according to priority until the end or the first occurence of multiple filter values in one filed,right? I think it's a little complex and hard to understand.

simozhan:
@ycheng-aa @lubomir yes. the original implementation is a bit complex. i changed it in the simple way, group rpm query parameters and not supporting multi value filter for them. Maybe if someone wants to use multi value filter, he can use rpm_nvr, rpm_nvra and rpm_* together.

pdc/apps/compose/views.py
```diff
@@ -468,35 +467,19 @@ def filter_queryset(self, qs):
...
                                        ('rpm_arch', 'arch')]
-        for query_param, rpm_key in query_parma_rpm_key_mapping:
-            qs, q = self._filter_rpm_value(query_params, query_param, rpm_key, qs, q)
-            if q is None:
-                return qs
+        for query_param, rpm_key in query_param_rpm_key_mapping:
+            rpm_value = query_params.get(query_param, None)
+            s = 'variant__variantarch__composerpm__rpm__' + rpm_key
+            if rpm_value:
+                q &= Q(**{s + '__iexact': rpm_value})
```
ycheng-aa: If user input multiple values for one field, it will return empty queryset, right? If so, should it be documented?

Another thing is if you already handled these fields here, should them be popped from `self.request.query_params` to avoid be handled in filter part again?

simozhan: if multiple values for one field, the last one will be used. i guess this is the QueryDict's default behavior. i think for all the query fields in drf will follow this, we don't need to document the specific param.
And the QueryDict instance is immutable. This means you cannot change their attributes directly.
actually for the filter_class, there are no rpm_* params, they will not be handled.

lubomir:
Code-wise it looks good to me. If the same filter is used multiple times, only the last value seems to be used. That would not be a problem by itself, but it should probably be documented.

Based on the logged SQL queries it seems to me it now applies all filters to the same RPM, so when name and version filters are used together, it filters composes that have the RPM in that particular version. That is much better than the original behaviour I think.

ycheng-aa:
If it's the drf default behavior, it's good for me.

:+1:

bliuredhat:
:+1:

simozhan:
@lubomir @ycheng-aa yes. you are right. Using the last value is the drf's default behaviour, but we have already make most of the query params with multi values which change the default behaviour. we need to let the user knows these params are different from the others.

bliuredhat:
:+1:

ycheng-aa:
:+1:

PR #384, opened 2016-06-24
Fix sync.sh problem in upstream version part.

ycheng-aa:
1. Define the productmd dependency to specific version.
2. Fix a module error when run script from internal
   code base.

JIRA: PDC-1572


simozhan:
looks good

PR #385, opened 2016-06-28
Fix bugs when inputing wrong format which will return server error

simozhan:
JIRA: PDC-1560

### DIFF
--- a/pdc/apps/compose/views.py
+++ b/pdc/apps/compose/views.py
@@ -606,6 +606,10 @@ class ComposeViewSet(StrictQueryParamMixin,
         if not request.data:
             return NoEmptyPatchMixin.make_response()
 
+        if not isinstance(request.data, dict):
+            return Response(data={"detail": ("The parameters' format for updating is wrong. "
+                                             "Please read API documentation")}, status=status.HTTP_400_BAD_REQUEST)
+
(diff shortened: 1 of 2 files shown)


pdc/apps/compose/views.py
```diff
@@ -993,7 +997,7 @@ def _update_parameters_acceptable(self, in_data, layer):
         if layer == 1:
             result = isinstance(in_data, list)
         elif layer > 1:
-            if not isinstance(in_data, dict) and len(in_data) != 1:
+            if not isinstance(in_data, dict):
```
ycheng-aa: Should we allow dict length greater than 1?

lubomir:
LGTM. The reason why the RPM mapping end-point was not very robust with regards to bad input was that when it was written, it was not clear whether it actually satisfies RCM requirements. As far as I know, there has been no use of it yet, so we still don't know.

pdc/apps/compose/views.py
simozhan: Empty dict shouldn't be allowed, will change it.

simozhan:
@ycheng-aa @lubomir Currently i changed it to not allow empty and duplicated dicts as input. If future they have further requirements when using this api, let's change it. Does it make sense?

ycheng-aa:
:+1:

lubomir:
@simozhan Makes perfect sense :+1:

PR #386, opened 2016-07-04
Add cache control headers to PDC HTTP response

ycheng-aa:
JIRA: PDC-1577

### DIFF
--- a/pdc/settings.py
+++ b/pdc/settings.py
@@ -33,6 +33,9 @@ DEBUG = False
 
 ALLOWED_HOSTS = []
 
+# The default number of seconds to cache a page for the cache middleware
+CACHE_MIDDLEWARE_SECONDS = 300
+
 ITEMS_PER_PAGE = 50
 
 # ======== resource permissions configuration =========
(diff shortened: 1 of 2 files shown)


bliuredhat:
:+1:

pdc/settings.py
```diff
@@ -118,11 +122,14 @@
     'pdc.apps.usage.middleware.UsageMiddleware',
     'pdc.apps.changeset.middleware.ChangesetMiddleware',
     'pdc.apps.utils.middleware.MessagingMiddleware',
-    'pdc.apps.utils.middleware.RestrictAdminMiddleware'
+    'pdc.apps.utils.middleware.RestrictAdminMiddleware',
+    'django.middleware.common.CommonMiddleware',
```
simozhan: CommonMiddleware is already on line 113

simozhan:
i don't have much experience with cache and i am following this guide https://docs.djangoproject.com/en/1.9/topics/cache/
It seems in this guide, a CACHES setting need to be added which i haven't seen. But your test runs ok. So is the CACHES necessary?

ycheng-aa:
I added CACHES item but removed it. The CACHES could let Django use it's owe cache instead of calculate again. This commit looks only include generating cache control header. And I am also afraid if use Django CACHES, client may be no way to get the newest data. I am not sure. I hope the first commit of cache could make less impact.

pdc/settings.py
ycheng-aa: Fixed in 13acb9cf5b4a4d3. Thanks for the careful review.

ycheng-aa:
And I will also abandon internal code part change. Because no need to generate cache table if don't set `CACHES` here.

simozhan:
looks good to me then.

PR #388, opened 2016-07-14
Implement more accurate last-modified header in some resources of PDC

ycheng-aa:
Make PDC response HTTP last-modified header according to
related model change recorded in changset.
First support it in releases, release-variants, products, arches.

JIRA: PDC-1624

### DIFF
--- a/pdc/apps/release/models.py
+++ b/pdc/apps/release/models.py
@@ -371,7 +371,7 @@ class VariantArch(models.Model):
         ordering = ("variant", "arch")
 
     def __unicode__(self):
-        return u"%s.%s" % (self.variant, self.arch)
+        return u"%s.%s, deleted: %s" % (self.variant, self.arch, self.deleted)
 
 
 class ReleaseGroupType(models.Model):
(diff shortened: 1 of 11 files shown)


bliuredhat:
:+1:

pdc/apps/release/models.py
```diff
@@ -371,7 +371,7 @@ class Meta:
         ordering = ("variant", "arch")
 
     def __unicode__(self):
-        return u"%s.%s" % (self.variant, self.arch)
+        return u"%s.%s, deleted: %s" % (self.variant, self.arch, self.deleted)
```
simozhan: i am not quite sure, is this line's fix related to this task?

simozhan:
Looks good to me. :+1:

pdc/apps/release/models.py
ycheng-aa: I just to want to if a model is changed the it's unicode could reflect that. Because changeset will check if original and modified obj export value is same, if same then will not record the change. And this unicode method will be refered by other model,  like `Variant`.

simozhan: ok. got it. sounds reasonable.

PR #394, opened 2016-07-20
Create release-interop-feature-categories new API

bliuredhat:
This API have two parameter:
        name: (required, lower-case, no space, unique string)
        description: (optional, string)
 With GET, POST, UPDATE,DELETE method.

JIRA: PDC-1617

### DIFF
--- a/pdc/apps/release/filters.py
+++ b/pdc/apps/release/filters.py
@@ -6,7 +6,8 @@
 import django_filters
 
 from pdc.apps.common import filters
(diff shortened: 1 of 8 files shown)


ycheng-aa:
Is the requirement clear? I see there is still question pending in JIRA. The answer looks may affect the implementation.

bliuredhat:
@ycheng-aa, I am not sure and will ask @sochotnicky .

PR #399, opened 2016-07-29
Take keys into acount for ordering of all APIs

bliuredhat:
1. Add the OrderingFilter class supports simple query parameter
   controlled ordering of results.
2. This feature is availbale for these APIs which have the 'List'
   method and have the filter attribute.
3. The value which ordering key mapped is contained in given
   serializer_class field.

JIRA: PDC-1512

### DIFF
--- a/pdc/apps/common/viewsets.py
+++ b/pdc/apps/common/viewsets.py
@@ -195,13 +195,21 @@ class StrictQueryParamMixin(object):
         extra_keys = set(request.query_params.keys()) - allowed_keys
         if extra_keys:
             raise FieldError('Unknown query params: %s.' % ', '.join(sorted(extra_keys)))
-
         # If 'ordering' in query parameter, check the key whether in fields.
         if 'ordering' in request.query_params.keys():
-            if request.query_params.get('ordering') not in self.__class__.serializer_class.Meta.fields:
+            self._check_ordering_keys(request)
+
+    def _check_ordering_keys(self, request):
+        ordering_keys = request.query_params.get('ordering')
+        tmp_list = []
+        if ordering_keys.startswith('-'):
+            tmp_list.append(ordering_keys.lstrip('-'))
+        else:
+            tmp_list += ordering_keys.split(',')
+        for key in tmp_list:
+            if key not in self.serializer_class.Meta.fields:
                 raise FieldError('Unknown query key: %s not in fields: %s' %
-                                 (request.query_params.get('ordering'),
-                                  self.__class__.serializer_class.Meta.fields))
+                                 (key, self.serializer_class.Meta.fields))
 
 
 class PermissionMixin(object):
--- a/pdc/apps/common/viewsets.py
+++ b/pdc/apps/common/viewsets.py
@@ -176,7 +176,12 @@ class StrictQueryParamMixin(object):
     `query_params` class attribute on the serializer. Note that this should
     only include the parameters that are passed via URL query string, not
     request body fields.
+    ordering_fields parameter is prepared for OrderingFilter class,the current
+    value of ordering_fields means user can override ordering the results
+    by given key.
     """
+    ordering_fields = '__all__'
+
     def initial(self, request, *args, **kwargs):
         super(StrictQueryParamMixin, self).initial(request, *args, **kwargs)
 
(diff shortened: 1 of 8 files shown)


simozhan:
I am guessing the ordering is decided by name and description together, it seems the description's ordering works, but name not.

simozhan:
BTW, can we create test case with release as it is the one that user mentions.

pdc/apps/common/viewsets.py
```diff
@@ -186,12 +191,18 @@ def initial(self, request, *args, **kwargs):
...
             return
-
         allowed_keys = drf_introspection.get_allowed_query_params(self)
         extra_keys = set(request.query_params.keys()) - allowed_keys
         if extra_keys:
             raise FieldError('Unknown query params: %s.' % ', '.join(sorted(extra_keys)))
 
+        # If 'ordering' in query parameter, check the key whether in fields.
+        if 'ordering' in request.query_params.keys():
+            if request.query_params.get('ordering') not in self.__class__.serializer_class.Meta.fields:
```
lubomir: Based on the documentation of `OrderingFilter`, it looks like the value can be a comma separated list of fields. That would not work with this check. Also, the field names can be prefixed with `-` to indicate reverse order, which would not work here either.

lubomir:
From my simple testing it seems to work ok. I think there are two more changes needed: the Release and Compose list views have in-code definitions of sorting, which overrides the filter from URL. Those should be fixed to only use the explicit ordering if the query parameter is missing.

bliuredhat:
@lubomir; thank you; maybe I need to design it again; your suggestion is very useful for me.

bliuredhat:
@lubomir, fixed it in  4a9d514.

bliuredhat:
Hi, @lubomir, @simozhan and @ycheng-aa; please help to review this patch.
The function of OrderingFilter, which will order the results correspond to the key and ignored the default ordering of the results which defined in model.

simozhan:
Is the change in this way:
If you want to order release in short, version,type, you need to explicitly set query parameter like ?ordering=short,version,type which will override the default ordering in model?

pdc/apps/common/viewsets.py
```diff
@@ -186,11 +191,25 @@ def initial(self, request, *args, **kwargs):
...
+        if 'ordering' in request.query_params.keys():
+            self._check_ordering_keys(request)
+
+    def _check_ordering_keys(self, request):
+        ordering_keys = request.query_params.get('ordering')
+        tmp_list = []
+        if ordering_keys.startswith('-'):
+            tmp_list.append(ordering_keys.lstrip('-'))
+        else:
+            tmp_list += ordering_keys.split(',')
```
ycheng-aa: Could `,` and `-` work together? If not the code is OK. If allow it, looks we should split by `,` first and check `-`.

pdc/apps/common/viewsets.py
```diff
@@ -186,11 +191,22 @@ def initial(self, request, *args, **kwargs):
...
+            self._check_ordering_keys(request)
+
+    def _check_ordering_keys(self, request):
+        ordering_keys = request.query_params.get('ordering')
+        tmp_list = [param.strip() for param in ordering_keys.split(',')]
+
+        for key in tmp_list:
+            if key.lstrip('-') not in self.serializer_class.Meta.fields:
+                raise FieldError('Unknown query key: %s not in fields: %s' %
+                                 (key.lstrip('-'), self.serializer_class.Meta.fields))
```
ycheng-aa: It will display the first illegal field, not all illegal fields. Maybe it could be something like:
```
if set(tmp_list) - set(self.serializer_class.Meta.fields):
     # raise error with all illegal fields.
     ....................
```

pdc/apps/common/viewsets.py
```diff
@@ -186,11 +191,22 @@ def initial(self, request, *args, **kwargs):
...
         extra_keys = set(request.query_params.keys()) - allowed_keys
         if extra_keys:
             raise FieldError('Unknown query params: %s.' % ', '.join(sorted(extra_keys)))
+        # If 'ordering' in query parameter, check the key whether in fields.
+        if 'ordering' in request.query_params.keys():
+            self._check_ordering_keys(request)
+
+    def _check_ordering_keys(self, request):
+        ordering_keys = request.query_params.get('ordering')
+        tmp_list = [param.strip() for param in ordering_keys.split(',')]
```
ycheng-aa: Change `param.strip()` to `param.strip().lstrip('-')` here then it don't have to use lstrip in following code.

pdc/apps/release/tests.py
```diff
@@ -1829,16 +1829,32 @@ def test_retrieve_with_name(self):
...
+
+        self.assertEquals(response.data.get('results')[0], expect_result)
+        response1 = self.client.get(reverse("releasegroups-list"), {'ordering': 'description'},
+                                    format='json')
+        expect_result1 = {'active': True, 'type': u'QuarterlyUpdate',
+                          'name': u'rhel_test1', 'releases': [u'release-1.0'],
+                          'description': u'good'}
+        self.assertEqual(response1.status_code, status.HTTP_200_OK)
+        self.assertEquals(response1.data.get('results')[0], expect_result1)
+
```
ycheng-aa: It's better to test multiple fields ordering and order with `-`. If we support use them together, it looks should be tested also.

bliuredhat:
@ycheng-aa ; thank you; and I have updated as your suggestion. Please help to review~!

simozhan:
LGTM

ycheng-aa:
:+1:

PR #401, opened 2016-08-03
Unnest group resource permissions create/update/response data.

ycheng-aa:
Use one layer data in creating, updating and response
instead of old two layers data in group resource permissions
API.
JIRA: PDC-1551


simozhan:
:+1:

bliuredhat:
looks good for me

PR #403, opened 2016-08-04
Update the valid fields for ordering

bliuredhat:
The valid fields is included in model fields of all APIs.

JIRA: PDC-1512

### DIFF
--- a/docs/source/using_api.rst
+++ b/docs/source/using_api.rst
@@ -114,3 +114,20 @@ it from the menu in top right corner.
 
 To store the reason for the change, add HTTP header ``PDC-Change-Comment``,
 whose value is an arbitrary string that will be stored with the change.
+
+
+Override Ordering
+-----------------
+
+The client can override the ordering of the results with query parameter.
+By default, the query parameter is named ``ordering``.
+For example, to order releases by release_id:
+
+http://example.com/rest_api/v1/releases/?ordering=release_id
+
+The client may also specify reverse orderings by prefixing the field name with '-'.
+
+For example, to reverse orderings of releases by release_id:
+
+http://example.com/rest_api/v1/releases/?ordering=-release_id
+
--- a/pdc/apps/common/viewsets.py
+++ b/pdc/apps/common/viewsets.py
@@ -201,11 +201,13 @@ class StrictQueryParamMixin(object):
 
     def _check_ordering_keys(self, request):
         ordering_keys = request.query_params.get('ordering')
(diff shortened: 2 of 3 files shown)


bliuredhat:
Document link: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/filters.py

simozhan:
Generally looks good to me. Just one thing to mention as we discussed, does it make sense to predefine all the ordering fields instead of using the fields in model, some field in model does not make much sense, like integrated_with in release model.

pdc/apps/common/viewsets.py
```diff
@@ -201,11 +201,13 @@ def initial(self, request, *args, **kwargs):
 
     def _check_ordering_keys(self, request):
         ordering_keys = request.query_params.get('ordering')
+        valid_fields = [field.name for field in self.queryset.model._meta.fields]
```
ycheng-aa: I think we discussed, and @simozhan also mentioned. Should it use model's fields? The reasons might be:

1. It could be different name between model and serializer. As defined in `source` of a serializer's field.
2. Serializer's fields could be more than model's. 

Is it still OK to use model's fields?

docs/source/using_api.rst
```diff
@@ -114,3 +114,23 @@ it from the menu in top right corner.
...
 whose value is an arbitrary string that will be stored with the change.
+
+
+Override Ordering
+-----------------
+
+With OrderingFilter class, the users can override the ordering of the results
+with query parameter. By default, the query parameter is named ``ordering``. And
+the ``ordering_fields``, which by using the special value ``__all__``, means that
+a view allow ordering on any ``model`` field or queryset aggregate.
```
ycheng-aa: I think customer should no care about what is `model`. It's internal implementation, customer also don't know what fields in a model. Also as discussed, should we use model's fields?

bliuredhat:
@ycheng-aa; For OderingFilter class, there 2 ways to realize this task: 
1. When don't specify an ordering_fields, the valid fields  is  serializer_fileds .
2. When  ordering_fields = '__all__',    the valid fields is model_fileds.
But in current, we define ordering_fields = '__all__', so the valid fields is  model_fileds.

BTW; I will push other patch which by using serializer_fileds as the valid fields.

ycheng-aa:
OK then. But the doc `a view allow ordering on any ``model`` field or queryset aggregate.` seems not for customer. It's more like for developer. Maybe we should not expose this detail to customer.

PR #404, opened 2016-08-05
Override the ordering with the serializer_fileds

bliuredhat:
In this patch; don't specify the ordering_fields attribute, so the valid fields is serializer_fileds.
JIRA: PDC-1512

### DIFF
--- a/pdc/apps/common/viewsets.py
+++ b/pdc/apps/common/viewsets.py
@@ -180,7 +180,6 @@ class StrictQueryParamMixin(object):
     value of ordering_fields means user can override ordering the results
     by given key.
     """
-    ordering_fields = '__all__'
 
     def initial(self, request, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


bliuredhat:
The document link: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/filters.py#L295 .

bliuredhat:
* The valid fileds from serializer_class:
 ['release_id', 'short', 'version', 'name', 'base_product', 'active', 'product_version', 'release_type', 'integrated_with', 'releasebugzillamapping', 'releasedistgitmapping']
* The valid fileds from Model_class:
[u'id', 'release_id', 'short', 'version', 'name', 'release_type', 'base_product', 'active', 'product_version', 'integrated_with']
There are difference between them; in my opinion; I prefer to choice Model_class filed for ordering.

Issue #319, opened 2016-03-22
Display all/active/inactive releases, release versions and products


dmach:
There should be a drop-down next to search field that allows user to select between all/active/inactive releases, product versions and products.
Default should be: active

bliuredhat:
@dmach; for the select `all`/`active`/`inactive`; I prefer to set the default value: `all`, not `active`. 
Do you agree?
Please help to add some comments here.

bliuredhat:
This issue have been finished. It have been deployed in dt server, and can check the webUI links as followed .
* https://pdc-dt.host.qe.eng.pek2.redhat.com/release/;
* https://pdc-dt.host.qe.eng.pek2.redhat.com/product/;
* https://pdc-dt.host.qe.eng.pek2.redhat.com/product-version/.

PR #412, opened 2016-08-26
Display all/active/inactive 

bliuredhat:
For releases, product versions and products APIs; add selector attribute
 in datatable, which instead of the Column Name.And click the seletor
 to filter and display all/active/inactive status.
JIRA: PDC-1589

### DIFF
--- a/pdc/apps/release/templates/product_list.html
+++ b/pdc/apps/release/templates/product_list.html
@@ -9,15 +9,13 @@
     {% include "help_button.html" %}
 </h2>
 
-<form method="GET" style="float: left;">
+<form method="GET" style="left">
   {% csrf_token %}
   {{ form.search }}
   <input type="submit" value="{% trans "Search" %}" />
 </form>
 
-{% include "pagination.html" %}
 {% include "product_list_include.html" %}
-{% include "pagination.html" %}
 
 {% helppopover %}
     <p>In PDC, a <em>product</em> is just a group of
--- a/pdc/apps/release/templates/product_list_include.html
+++ b/pdc/apps/release/templates/product_list_include.html
@@ -1,10 +1,16 @@
 {% load i18n %}
-<table class="table table-striped">
+<table class="table table-striped table-bordered">
  <thead>
   <tr>
     <th>{% trans "Short Name" %}</th>
     <th>{% trans "Name" %}</th>
-    <th>{% trans "Active" %}</th>
+    <th>
+        <select id="active_filter" value= {{paginator.per_page}}>
+            <option value="All" selected>all</option>
+            <option value="True">active</option>
+            <option value="False">inactive</option>
+        </select>
+    </th>
     <th>{% trans "Product Versions [active/total]" %}</th>
     <th>{% trans "Releases [active/total]" %}</th>
   </tr>
--- a/pdc/static/js/datatable_active_filter.js
+++ b/pdc/static/js/datatable_active_filter.js
@@ -1,0 +1,22 @@
+(function () {
+    "use strict";
+    $(document).ready(function() {
+        if ($('select').is('#active_filter')){
+            var page_size = parseInt($('select').attr("value"));
+            var table = $('table').dataTable(
+                {   "pageLength" : page_size,
+                    "ordering": false
+                });
+            $('.dataTables_filter').hide();
+            $('#active_filter').prop('selectedIndex',0);
+            $('#active_filter').on('change', function() {
+                if (this.value !== "All") {
+                    table.fnFilter( $(this).val());
+                }
+                else {
+                    table.fnFilter("");
+                }
+            });
+        }
+    });
+})();
--- a/pdc/apps/release/templates/product_version_list.html
+++ b/pdc/apps/release/templates/product_version_list.html
@@ -9,15 +9,13 @@
     {% include "help_button.html" %}
 </h2>
 
-<form method="GET" style="float: left;">
+<form method="GET" style="left">
   {% csrf_token %}
   {{ form.search }}
   <input type="submit" value="{% trans "Search" %}" />
 </form>
 
-{% include "pagination.html" %}
 {% include "product_version_list_include.html" %}
-{% include "pagination.html" %}
 
 {% helppopover %}
     <p>In PDC, a <em>product version</em> is just a group of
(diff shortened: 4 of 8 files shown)


ycheng-aa:
I tested it in Firefox and Chrome, in Firefox it works as the code logic, but in Chrome it didn't. I am not sure it's my setting problem or some other problems. Did you also test in Chrome?

The logic looks like it will show all hide based on current page content. If current page don't have `inactive`  releases, and user select `inactive`, then it will show a page with no release content, even if there are inactive releases in DB. Is it right? If so, it seems pretty inconvenient for customer.

lubomir:
Minor nitpick: the script should not be copied in three places. Either is should be included via `src` attribute or at Django template level.

bliuredhat:
@lubomir ; thank you; your ideal is right; and I will update it.

bliuredhat:
@ycheng-aa ; For the first  question, I can't find the reason why can't work in Chrome; maybe we need to check it in detail.
For  the second question; yes, the current patch can't meet the requirements of this feature; I will update it.

lubomir:
Overall it seems to be working correctly. However, I'm not a great fan of the whole interaction: when the page is loaded, all items are displayed. I select only active, and suddenly the whole table jumps down, because the datatable header appears. Select All again and the table jumps back up.

Maybe the table could always be displayed as a datatable? (It would require disabling sorting by the activity column as well.)
Is it necessary to reload the page when setting the filter to All?

I also think the labels should be improved: when you select only active, it will display a column where all values are True and the header also says True. It is not obvious what it actually means. The only other way I can think of is *Active*, *Only active* and *Only inactive*, which is not much better. Do you know any place that uses this design pattern? Maybe UX team could help us out here?

The problem that it only filters based on the current page could be fixed by displaying all objects and doing the pagination completely with datatable. Some testing would be needed to see how well it works with a lot of rows.

bliuredhat:
@lubomir ;thank you, I am agree with you. I will communicate with @ycheng-aa,  about "the problem that it only filters based on the current page could be fixed by displaying all objects and doing the pagination completely with datatable. "

pdc/static/js/datatable_active_filter.js
```diff
@@ -0,0 +1,15 @@
...
+            if (this.value !== "All") {
+                table.fnFilter( $(this).val()); 
+                $('.dataTables_filter').hide();
+            }
+            else { 
+                location.reload();
+                 }
+             }); 
+    });
+})();
```
ycheng-aa: Add new line.

pdc/static/js/datatable_active_filter.js
```diff
@@ -0,0 +1,15 @@
...
+    "use strict";
+    $(document).ready(function() {
+        $('#active_filter').on('change', function() {
+            var table = $('table').dataTable();
+            if (this.value !== "All") {
+                table.fnFilter( $(this).val()); 
+                $('.dataTables_filter').hide();
+            }
+            else { 
+                location.reload();
```
ycheng-aa: Why reload only select 'all'?

ycheng-aa:
1. There are different styles when display 'all' and 'active/inactive'. Looks inconsistent. E.g. number of items in one page.  50 for `all` and 20 for `active/inactive`.

2. The current implementation looks still filter based current page when select 'active/inactive'. 'all' results has it's own pagination, 'active/inactive' filtering result pagination further based on current page of 'all' result. Is my understanding correct?  If so, I don't think it should be like that. For example, if the only one `inactive` release is displayed in the last page. User should first go to the last page, and select `inactive`. This inactive release can't be found by filtering `inactive` in the first page, the second page, the third page, ...., until the last page.
Agree with @lubomir. I think the solution get all objects and do all filter and pagination by datatable should work. But I don't have too much experience. I can't give more detailed suggestion.

pdc/static/js/datatable_active_filter.js
bliuredhat: I will update with "table.fnFilterClear(); "

bliuredhat:
@lubomir @ycheng-aa @simozhan ; some changes as followed;
1, Remove the pagination once we used, and apply the Datatables, which reference to jQuery and can complete pagination automatically.
2. The new pagination also can set the page_size in setting file.

Please help to review this patch.

simozhan:
And another thing is when this feature is ready for merge, please squash the four commits in one before merge.

bliuredhat:
@simozhan ; this suggestion is from @lubomir. 
BTW, @ycheng-aa and @simozhan; if we have a good choice, I will update it later; please add your suggestion here. Think you~!

pdc/apps/release/templates/product_list_include.html
```diff
@@ -5,10 +5,10 @@
     <th>{% trans "Short Name" %}</th>
     <th>{% trans "Name" %}</th>
     <th>
-        <select id="active_filter">
-            <option value="All">Active-ALL</option>
-            <option value="True">True</option>
-            <option value="False">False</option>
+        <select id="active_filter" value= {{paginator.per_page}}>
```
lubomir: What is the value here used for?

pdc/apps/release/templates/product_list.html
```diff
@@ -9,7 +9,7 @@ <h2 class="page-header">
     {% include "help_button.html" %}
 </h2>
 
-<form method="GET" style="float: left;">
+<form method="GET" style="show float: left;">
```
lubomir: Wouldn't removing the `float` attribute work just as well?

lubomir:
@simozhan I really don't know how to change the labels to be clear: I think it would be nice if it made sense as a column header. But ultimately let's not block on such detail; it can always be changed.

Functionality wise, the latest version seem good to me. Maybe adding `table-bordered` class would be nice to be consistent with datatables for repos (on release detail page).

pdc/apps/release/templates/product_list.html
bliuredhat: yes, it work well when remove `show float` .

pdc/apps/release/templates/product_list_include.html
bliuredhat: @lubomir; it don't work well when we apply `data-per-page`, because the Datatables will override the pagination, and set the page_size as 20.

bliuredhat:
@lubomir , when adding table-bordered class in table, it looks more beautiful.

bliuredhat:
@all; Update the drop-down list as  all/active/inactive, which according to the requirement: https://github.com/product-definition-center/product-definition-center/issues/319

simozhan:
:+1:

PR #414, opened 2016-09-05
Update the docs and add the ordering fields

bliuredhat:
When override the ordering of the results, some changes happened;
as followed:
1, Update the ordering fields. the ordering fields are the intersection
of Model fields and Serializer fields.
2. To make the ordering fields more friendly and clearly for customers,
add the ordering fields in docs, which share in page of 'LIST' method.

JIRA:PDC-1659

### DIFF
--- a/pdc/apps/common/renderers.py
+++ b/pdc/apps/common/renderers.py
@@ -65,6 +65,7 @@ Responses are available in JSON format.
 """
 
 URL_SPEC_RE = re.compile(r'\$(?P<type>URL|LINK):(?P<details>[^$]+)\$')
+ORDERING_STRING = "\n * `ordering` is used to override the ordering of the results, the value could be: "
 
 
 class ReadOnlyBrowsableAPIRenderer(BrowsableAPIRenderer):
@@ -140,6 +141,11 @@ class ReadOnlyBrowsableAPIRenderer(BrowsableAPIRenderer):
         macros = settings.BROWSABLE_DOCUMENT_MACROS
         if view:
             macros['FILTERS'] = get_filters(view)
+            if 'list' == method:
+                ordering_field = get_ordering_field(view, method)
+                if ordering_field:
+                    ordering_string = ORDERING_STRING + " %s ." % ordering_field
+                    macros['FILTERS'] += ordering_string
             if '%(SERIALIZER)s' in docstring:
                 macros['SERIALIZER'] = get_serializer(view, include_read_only=True)
             if '%(WRITABLE_SERIALIZER)s' in docstring:
@@ -172,6 +178,22 @@ class ReadOnlyBrowsableAPIRenderer(BrowsableAPIRenderer):
                              exc_info=sys.exc_info())
                 return 'BAD URL'
         return URL_SPEC_RE.sub(replace_url, text)
+
+
+def get_ordering_field(view, method):
+    """ If the APIs have the LIST method; for the view of LIST method, add the
+    Ordering field for the users.
+    """
+    if 'list' in method and view.serializer_class:
+        model_fields = [field.name for field in view.queryset.model._meta.fields]
+        serializer_fields = [
+            field.source or field_name
+            for field_name, field in view.serializer_class().fields.items()
+            if not getattr(field, 'write_only', False)]
+        valid_fields = list(set(model_fields).intersection(set(serializer_fields)))
+        return valid_fields
+    else:
+        return None
 
 
 FILTERS_CACHE = {}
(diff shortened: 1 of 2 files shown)


pdc/apps/common/renderers.py
```diff
@@ -65,6 +65,8 @@
 """
 
 URL_SPEC_RE = re.compile(r'\$(?P<type>URL|LINK):(?P<details>[^$]+)\$')
+ORDERING_STRING = "\n \n * __Ordering Field__ is used to override the ordering of the results by client, " \
```
simozhan: is it ok to change Ordering Field to ordering and make it the same style/format with other query params

simozhan: is it better to move by client, i am not sure.

bliuredhat:
@simozhan ; I consider that it's not a query params here, it's the docs and the aim is to inform ordering field message.

simozhan:
looks good to me

pdc/apps/common/renderers.py
```diff
@@ -65,6 +65,7 @@
 """
 
 URL_SPEC_RE = re.compile(r'\$(?P<type>URL|LINK):(?P<details>[^$]+)\$')
+ORDERING_STRING = "\n * `ordering` is used to override the ordering of the results, the value is: "
```
ycheng-aa: Maybe could use 'the value could be:`?

pdc/apps/common/renderers.py
```diff
@@ -140,6 +141,11 @@ def format_docstring(self, view, method, docstring):
         macros = settings.BROWSABLE_DOCUMENT_MACROS
         if view:
             macros['FILTERS'] = get_filters(view)
+            if 'list' == method:
+                ordering_field = get_ordering_field(view, method)
+                if ordering_field:
+                    ordering_string = ORDERING_STRING + "\n %s ." % ordering_field
```
ycheng-aa: It is better to add markdown format. E.g surround each field with `` ` `` could make it read, etc. And it looks same with or without above `\n`.

ycheng-aa: It is better to add markdown format. E.g surround each field with `` ` `` could make it read, etc. And it looks same with or without above `\n`.

pdc/apps/common/renderers.py
```diff
@@ -174,6 +180,20 @@ def replace_url(match):
         return URL_SPEC_RE.sub(replace_url, text)
 
 
+def get_ordering_field(view, method):
+    """ If the APIs have the LIST method; for the view of LIST method, add the
+    Ordering field for the users.
+    """
+    if 'list' in method and view.serializer_class:
+        model_fields = [field.name for field in view.queryset.model._meta.fields]
```
ycheng-aa: It will not meet a `return` statement if the `if` result is false.

pdc/apps/common/renderers.py
```diff
@@ -174,6 +180,20 @@ def replace_url(match):
         return URL_SPEC_RE.sub(replace_url, text)
 
 
+def get_ordering_field(view, method):
+    """ If the APIs have the LIST method; for the view of LIST method, add the
+    Ordering field for the users.
+    """
+    if 'list' in method and view.serializer_class:
```
ycheng-aa: It will not meet a return statement if the if result is false.

bliuredhat: it returns None, and then, it will don't update the docs in [Query params] parts.

pdc/apps/common/renderers.py
bliuredhat: yes, I remove the `\n` here.

bliuredhat:
@all, thank you for help to review.

pdc/apps/common/renderers.py
```diff
@@ -140,6 +141,11 @@ def format_docstring(self, view, method, docstring):
         macros = settings.BROWSABLE_DOCUMENT_MACROS
         if view:
             macros['FILTERS'] = get_filters(view)
+            if 'list' == method:
+                ordering_field = get_ordering_field(view, method)
+                if ordering_field:
+                    ordering_string = ORDERING_STRING + " %s ." % ordering_field
```
ycheng-aa: Is it OK no markdown format here? For me, if the fields is read seems more obvious. I am not sure.

pdc/apps/common/renderers.py
ycheng-aa: Some people suggest it's better to have a explicit return if the return value is useful, E.g. http://stackoverflow.com/questions/4027586/is-it-ok-to-skip-return-none. For me it seems a little more readable. 
@simozhan what do you think?

bliuredhat: @ycheng-aa , okay; update it.

simozhan:
@bliuredhat i think @ycheng-aa is right. explicitly return a value is readable for human-beings

pdc/apps/common/renderers.py
bliuredhat: yes, it's work well.

PR #415, opened 2016-09-06
Change contact APIs' response max-age to 0

ycheng-aa:
Currently PDC set max-age to 30 based on the
assumption that PDC data doesn't change
frequently, and normally client can tolerate
maximum 30s delay.
But contact APIs are also accessed by contact
browser, it may change contacts frequently and
need result be reflected immediately in WEB.
So change max-age of global component contacts
and release component contacts API to 0.

JIRA: PDC-1699


simozhan:
looks good to me

bliuredhat:
:+1:

PR #416, opened 2016-09-07
Display all/active/inactive and filter the datatable

bliuredhat:
Fix the bug from QE

JIRA:PDC-1589

### DIFF
--- a/pdc/apps/release/views.py
+++ b/pdc/apps/release/views.py
@@ -35,13 +35,19 @@ from pdc.apps.auth.permissions import APIPermission
 from . import lib
 
 
-class ReleaseListView(SearchView):
+class PageSizeMixin(object):
+    def get_context_data(self, **kwargs):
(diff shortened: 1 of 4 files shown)


simozhan:
LGTM

pdc/apps/release/views.py
```diff
@@ -562,7 +570,11 @@ class ProductVersionListView(SearchView):
     allow_empty = True
     template_name = "product_version_list.html"
     context_object_name = "product_version_list"
-    paginate_by = settings.ITEMS_PER_PAGE
+
+    def get_context_data(self, **kwargs):
```
ycheng-aa: The above `get_context_data` almost same. Could implement it in a mixin and let each view inherit the mixin? Or some other way to eliminate duplicate code.

ycheng-aa:
Generally looks to me. It's better to eliminate duplicate code.

ycheng-aa:
:+1:

PR #418, opened 2016-09-12
Use slug in release / product version / product URLs.

dmach:
When using multiple PDC instances, it's difficult to preserve URLs,
since the IDs for a release/productversion/product can differ.
Using slug field solves the problem.


simozhan:
Yes. ID is meaningless between different PDC instances. :+1:

PR #420, opened 2016-09-13
Rename the colunm name for release, product version and product

bliuredhat:
JIRA: PDC-1589

### DIFF
--- a/pdc/apps/release/templates/product_list_include.html
+++ b/pdc/apps/release/templates/product_list_include.html
@@ -6,9 +6,9 @@
     <th>{% trans "Name" %}</th>
     <th>
         <select id="active_filter" value= {{ page_size }}>
-            <option value="All" selected>all</option>
-            <option value="True">active</option>
-            <option value="False">inactive</option>
+            <option value="All" selected>All products</option>
+            <option value="Active">Active products</option>
+            <option value="Inactive">Inactive products</option>
         </select>
     </th>
     <th>{% trans "Product Versions [active/total]" %}</th>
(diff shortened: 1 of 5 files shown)


ycheng-aa:
For me, the `True` / `False` is still not that clear if I first see the page. True for what and False for what. I don't have UI experience. As a user, it may still cause some confusion. Should it at least tell user what is True/False for?

bliuredhat:
@ycheng-aa ; I am confused. Do you mean that we change `True` value to `Active releases` in table ?

ycheng-aa:
@bliuredhat honestly I don't know how to change it. What I meant is, e.g, when you select `Inactive releases`, this column's value is False, it mean active status is false, right? But if a user don't have this backgroud, how should he know `False` mean active status is false. Nowhere says the `False` stand for active status is false.

bliuredhat:
@ycheng-aa , I see. your means is to use other string instead `True` or `False` in table. 
But I want to update the column name and make it more meaningful.

bliuredhat:
@ycheng-aa, @simozhan ; please help to review this patch. For this task, you maybe check the comments: https://projects.engineering.redhat.com/browse/PDC-1589.

ycheng-aa:
I still have the same problem, does it easy to understand what True/False mean in this column?  When I select 'Inactive releases`, this value is false. I am not sure it's easy to understand what does 'False' stand for.

Maybe you could invite more experienced people to review this.  I am ok if you get 2 `+1` to merge it.

simozhan:
@bliuredhat i don't have strong opinion with the column name. One issue i meet is when i click one product version for example rhel-7 and navigate to product version detail page, it seems no releases in that page, actually  there are several releases for that product version. So please help to check if this fix makes the regression.

bliuredhat:
@ycheng-aa ; I think you can't consider  in that ways, because this is a filter; and the function is to filter the mapping results from table. That is to say, if the column name is `False`,  the purpose is to filter the results with False value; this is the different between us.
Do you agree with me ?

ycheng-aa:
@bliuredhat  I think currently it is when you select 'Inactive releases`, each line in this column's value is `false`. Inactive is not false.  I am not sure if it's good.  Maybe your are right. So I think you can invite UI guys to review it. It's ok for me if they think it's fine.

bliuredhat:
@ycheng-aa , Thank you. And it's maybe need @sochotnicky to verify it.

bliuredhat:
@ycheng-aa @simozhan; I have updated as your suggestions; please help to review again~! thank you~!

ycheng-aa:
Looks good.

PR #422, opened 2016-09-28
Make release components be unique by name/release/type.

ralphbean:
This used to consider only `name` and `release`, which makes a certain
amount of sense.  We don't want more than one rpm with the same name in
the same release entry here.

However, if we open this up to start handling other types of release
components like `container` or `module`, then it is easy to imagine that
we might have one "cockpit" rpm in a release but also a "cockpit"
container in the same release.

If you try to create both of those entries in PDC, it will fail saying
that "name and release must be unique".  This change fixes that by
loosening the uniqueness constraint to also take `type` into account.

### DIFF
--- a/pdc/apps/component/migrations/0012_auto_20160928_1838.py
+++ b/pdc/apps/component/migrations/0012_auto_20160928_1838.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('component', '0011_auto_20151126_0602'),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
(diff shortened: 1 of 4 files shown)


ralphbean:
Ooop, this needs a migration added too to change the db constraint.

ralphbean:
OK, good to go now.

ycheng-aa:
:+1:

ralphbean:
> tests should be fixed

Good call.  Looking at this now.

ralphbean:
OK @ycheng-aa.  It should be all fixed up now.

ycheng-aa:
:+1:

bliuredhat:
Looks good for me

simozhan:
:+1:

ralphbean:
OK to merge this now?

bliuredhat:
@ralphbean , yes, please help to merge this patch.

ycheng-aa:
Merged.

PR #423, opened 2016-11-21
To list all API permissions

bliuredhat:
The WebUI shows the API permission, which includes all users and groups.
Because 'read' permission is available for all users and groups, instead of the values by @all.

JIRA: PDC-1757

### DIFF
--- a/examples/find-duplicated-contacts.py
+++ b/examples/find-duplicated-contacts.py
@@ -31,5 +31,6 @@ def run(resource):
             seen_roles.add(contact['contact_role'])
     print ''
 
+
 run('global-components')
 run('release-components')
--- a/pdc/apps/auth/views.py
+++ b/pdc/apps/auth/views.py
@@ -7,13 +7,15 @@
 import inspect
 import json
 
-from django.shortcuts import render, redirect
+from collections import OrderedDict
+from django.shortcuts import render, redirect, get_list_or_404
 from django.contrib.auth import (REDIRECT_FIELD_NAME, get_user_model,
                                  load_backend)
 from django.contrib.auth import logout as auth_logout
 from django.contrib.auth.models import Group, Permission
-from django.http import HttpResponseRedirect, HttpResponse
+from django.http import HttpResponseRedirect, HttpResponse, Http404
 from django.conf import settings
+from django.core.urlresolvers import NoReverseMatch
 
 from rest_framework import status
 from rest_framework.authtoken.models import Token
@@ -21,6 +23,7 @@ from rest_framework.decorators import list_route
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework import mixins, viewsets
+from rest_framework.reverse import reverse
 
 from . import backends
 from . import filters
@@ -30,8 +33,11 @@ from pdc.apps.auth.models import ResourcePermission, ActionPermission, Resource
 from pdc.apps.auth.permissions import APIPermission
 from pdc.apps.common.viewsets import StrictQueryParamMixin, ChangeSetUpdateModelMixin
 from pdc.apps.common import viewsets as common_viewsets
-from pdc.apps.utils.SortedRouter import router
-from pdc.apps.utils.utils import group_obj_export, convert_method_to_action, read_permission_for_all
+from pdc.apps.utils.SortedRouter import router, URL_ARG_RE
+from pdc.apps.utils.utils import (group_obj_export,
+                                  convert_method_to_action,
+                                  read_permission_for_all,
+                                  urldecode)
 
 
 def remoteuserlogin(request):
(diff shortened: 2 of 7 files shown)


bliuredhat:
@ycheng-aa and @simozhan ;
Please help to review this patch, thank you~!

examples/find-duplicated-contacts.py
```diff
@@ -31,5 +31,6 @@ def run(resource):
             seen_roles.add(contact['contact_role'])
     print ''
 
+
```
simozhan: a typo?

pdc/apps/common/renderers.py
```diff
@@ -248,6 +248,7 @@ def get_filters(view):
     FILTERS_CACHE[view] = filters
     return filters
 
+
```
simozhan: do we really need this blank line

bliuredhat: yes, if we don't add the blank line; can't pass when run  "make flake8" in github. https://travis-ci.org/product-definition-center/product-definition-center/builds/177633056 .

examples/find-duplicated-contacts.py
bliuredhat: yes

pdc/apps/auth/views.py
```diff
@@ -134,6 +140,142 @@ def user_profile(request):
...
+        pass
+
+    # get all users
+    users_list = []
+    users = models.User.objects.all()
+    for user in users:
+        if user.is_superuser:
+            users_list.append(str(user.username))
+        else:
+            resource_permission_set = get_resource_permission_set(user)
```
ycheng-aa: `get_resource_permission_set` also contain `is_superuser` judgement. Seems all could use logic here.

Besides, it looks pretty low efficiency. Even these users is gotten from groups, still need a loop for all users, maybe thousands times? Not sure. And just for one resource, and `get_resource_permission_set` also need to access DB several times. It looks very inefficient. I doubt if it could work well in real environment. 

Should we consider a non-super user get permissions out of his group now?

pdc/apps/auth/views.py
```diff
@@ -134,6 +140,142 @@ def user_profile(request):
...
+    api_root_dict = OrderedDict()
+    viewsets = {}
+    ret = OrderedDict()
+    list_name = router.routes[0].name
+    for prefix, viewset, basename in router.registry:
+        api_root_dict[prefix] = list_name.format(basename=basename)
+        viewsets[prefix] = viewset
+    sorted_api_root_dict = OrderedDict(sorted(api_root_dict.items()))
+    for key, url_name in sorted_api_root_dict.items():
+
```
ycheng-aa: Maybe remove this blank line looks better. Not sure.

pdc/apps/auth/views.py
```diff
@@ -134,6 +140,142 @@ def user_profile(request):
...
+    for obj in models.ResourcePermission.objects.all():
+
+        name = URL_ARG_RE.sub(r'{\1}', obj.resource.name)
+        url = ret[name]
+        members = get_users_and_groups(obj)
+        if members:
+            result = _dict_to_str(members)
+        else:
+            result = 'N/A'
```
ycheng-aa: Maybe should avoid use magic word in method. Use a constant maybe better.

pdc/apps/auth/views.py
bliuredhat: @ycheng-aa ; your suggestion is right, and  the current code is low efficiency.

pdc/apps/auth/views.py
```diff
@@ -140,15 +140,14 @@ def user_profile(request):
     return render(request, 'user_profile.html', context)
 
 
-def get_users_and_groups(arg):
+def get_users_and_groups(resource_permission):
     # get all groups
     members = dict()
     groups_list = []
```
ycheng-aa: this line seems useless.

pdc/apps/auth/views.py
```diff
@@ -159,8 +158,15 @@ def get_users_and_groups(arg):
...
-            if arg in resource_permission_set:
+            if read_permission_for_all():
+                resource_permission_set = set([obj for obj in models.ResourcePermission.objects.filter(
+                    permission__name__iexact='read')])
+            group_id_list = [group.id for group in user.groups.all()]
+            queryset = models.GroupResourcePermission.objects.filter(group__id__in=group_id_list)
+
+            for group_resource_permission in queryset:
+                resource_permission_set.add(group_resource_permission.resource_permission)
+            if resource_permission in resource_permission_set:
```
ycheng-aa: Seems don't have too much different. Still low efficiency and lot of duplicate code of `get_resource_permission_set`. Actual logic part also has some duplication. Already got group list have this resource permission before, here check user -> groups -> resource permissions again.

And again, Should we consider a non-super user get permissions out of his group now?

It would be nice to respond to every comment and question. No matter just something like `fixed` or some other discussion.

dmach:
Can you cherry-pick following commit?
It simplifies perm listing.
https://github.com/dmach/product-definition-center/commit/feb24588b610a1f14ff3183acf013dd2056fcd38

dmach:
Bugs/RFEs:
* all URLs point to /rest_api/perms - fix to point to the resources
* handle admins differently
  * display only people who have explicit perms
  * below the list, create another one simply listing people with admin privileges

pdc/apps/auth/views.py
```diff
@@ -134,6 +140,128 @@ def user_profile(request):
...
+        pass
+
+    # get all users
+    users_list = []
+    users = models.User.objects.all()
+    for user in users:
+        uer_group_list = [group.name for group in user.groups.all()]
+        if list(set(groups_list).intersection(set(uer_group_list))):
+            users_list.append(str(user.username))
+
```
ycheng-aa: Is above lines from line 155 same as something like following? Please help me check it,
```
users_set = { str(user.username) for user in get_user_model().objects.filter(groups__name__in=groups_list)}
users_list = list(users_set)
```

I think current implementation doesn't improve too much. Need to access DB number of users times.

pdc/apps/auth/views.py
```diff
@@ -152,12 +150,8 @@ def get_users_and_groups(resource_permission):
         pass
 
     # get all users
-    users_list = []
-    users = models.User.objects.all()
-    for user in users:
-        uer_group_list = [group.name for group in user.groups.all()]
-        if list(set(groups_list).intersection(set(uer_group_list))):
-            users_list.append(str(user.username))
+    users_set = {str(user.username) for user in get_user_model().objects.filter(groups__name__in=groups_list)}
```
ycheng-aa: Need to test it to make sure it works correctly. I didn't test it. Is this what you need in logic?

PR #424, opened 2016-11-23
Create a new API of api/resource-permission

bliuredhat:
This API is to show the permisssions of a resource, which includes
all users and groups.

1. This API support the list/retrieve method.
2. Because the User's permission depend on their Groups's permission,
only Group's permission can be changed.

JIRA: PDC-1756


pdc/apps/auth/views.py
```diff
@@ -903,3 +911,204 @@ def destroy(self, request, *args, **kwargs):
...
+                                   ChangeSetDestroyModelMixin,
+                                   mixins.RetrieveModelMixin,
+                                   mixins.ListModelMixin,
+                                   MultiLookupFieldMixin,
+                                   viewsets.GenericViewSet):
+    """
+    This end-point provides API resource permissions.
+    """
+    queryset = models.GroupResourcePermission.objects.all()
+    serializer_class = serializers.GroupResourcePermissionSerializer
```
ycheng-aa: Should it use a new serializer and implement your logic in that serializer? 

Should this API support create method? It looks duplicate of group resource permission API.

pdc/apps/auth/views.py
```diff
@@ -903,3 +911,204 @@ def destroy(self, request, *args, **kwargs):
...
+                                "users": list,
+                                "groups" list
+                            }
+                             ...
+                    }
+                    ...
+            }
+
+        """
+        result = {}
```
ycheng-aa: I don't quite understand why implement it here rather than in serializer. I am not quite experienced about code structure. But if you implement it here need to implement a lot of work serializer will do automatically, and I am not sure it support pagination, ordering, etc well.

pdc/apps/auth/views.py
```diff
@@ -903,3 +911,204 @@ def destroy(self, request, *args, **kwargs):
...
+            }
+
+        __Info__:
+
+        Only the permission of Group can be created.
+        """
+        return super(APIResourcePermissionViewSet, self).create(request, *args, **kwargs)
+
+    def destroy(self, request, *args, **kwargs):
+        """
```
ycheng-aa: Should this API support destory? Then `create` and `destroy` are all duplicate with old group resource permission API.

pdc/apps/auth/views.py
```diff
@@ -231,7 +237,9 @@ def get_api_perms(request):
     for obj in models.ResourcePermission.objects.all():
         name = URL_ARG_RE.sub(r'{\1}', obj.resource.name)
         url = ret[name]
-        members = get_users_and_groups(obj)
+        members = get_members(obj)
+        if read_permission_for_all() and obj.permission.name == 'read':
+            members = ['@all']
         perms.setdefault(name, OrderedDict()).setdefault(obj.permission.name, set()).update(members)
```
ycheng-aa: About part is not related to `Create a new API of api/resource-permission`. Normally in a separate commit is better.

pdc/apps/auth/views.py
bliuredhat: fixed it, 1fa7aefacb143d5da3266a5238833f2a0cfe611b

bliuredhat:
@ycheng-aa  and @simozhan ;
please help to review this patch , thank you.

pdc/apps/auth/tests.py
```diff
@@ -797,3 +797,96 @@ def test_list(self):
...
+        if hasattr(settings, 'DISABLE_RESOURCE_PERMISSION_CHECK'):
+            settings.DISABLE_RESOURCE_PERMISSION_CHECK = self.DISABLE_RESOURCE_PERMISSION_CHECK
+
+    def test_list(self):
+        url = reverse('apiresourcepermissions-list')
+        response = self.client.get(url, format='json')
+        self.assertEqual(len(response.data), 4)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_filter(self):
```
simozhan: This test method is called test_filter, but it seems like a list method, no filter included.

pdc/apps/auth/views.py
```diff
@@ -903,3 +910,231 @@ def destroy(self, request, *args, **kwargs):
...
+        allowed_keys = ['group', 'resource', 'permission']
+        data = request.data
+        if data:
+            extra_keys = set(data.keys()) - set(allowed_keys)
+            StrictSerializerMixin.maybe_raise_error(extra_keys)
+        group_data = data.get('group')
+        resource_data = data.get('resource')
+        permission_data = data.get('permission')
+
+        # TODO: verify the resource permission whether existed
```
simozhan: What does it mean TODO? something working in progress?

pdc/apps/auth/views.py
```diff
@@ -903,3 +910,231 @@ def destroy(self, request, *args, **kwargs):
...
+    queryset = models.ResourcePermission.objects.all()
+    serializer_class = serializers.ResourcePermissionSerializer
+    permission_classes = (APIPermission,)
+    lookup_fields = (
+        ('resource__name', r'[^/]+'),
+        ('permission__name', r'[^/]+'),
+    )
+    filter_class = filters.ResourcePermissionFilter
+
+    def _get_users_and_groups(self, resource_permission):
```
ycheng-aa: This method is nearly same as `get_users_and_groups`. It's not good just copy/paste to make duplicate code. E.g. if you fix read permission part later, you need to make change in 2 places rather than one. Implement it in one place make  UI  automatically consistent with API.

pdc/apps/auth/views.py
```diff
@@ -903,3 +910,231 @@ def destroy(self, request, *args, **kwargs):
...
+            pass
+        # get all users
+        superusers_set = {user.username for user in get_user_model().objects.filter(is_superuser=True)}
+        users_set = {user.username for user in get_user_model().objects.filter(groups__name__in=groups_list)}
+        users_list = list(superusers_set.union(users_set))
+        members['groups'] = sorted(groups_list)
+        members['users'] = sorted(users_list)
+        return members
+
+    def list(self, request, *args, **kwargs):
```
ycheng-aa: Still the same question, why don't you define your ow serializer and implement the logic in it. At first look, implement in that way seems more convenient, eg. it will take care of pagination automatically, make retrieve and list's element have same format.

Another thing I can't quite understand is why retrieve result has different format with list element's result. So customer will parse list result and retrieve result with totally different code. Does customer expect that?

pdc/apps/auth/views.py
ycheng-aa: OK. I asked @dmach a question in JIRA. If their functions and parameters are totally same, I don't quite understand the reason we do it again.

pdc/apps/auth/views.py
bliuredhat: @ycheng-aa ; I don't agree to your ideal; because this resource don't have model by itself, and in meta class, which need to add a model, this is a problem. For example, If we add {{ResourcePermission}} or {{GroupResourcePermision}} model, they are all not meet the requirement; in {{GroupResourcePermision}}, there is no {{user}} key, and {{ResourcePermission}} is too.
For the "retrieve result", I will update it.
BTW,  If my opinion is wrong, I will change it.

PR #426, opened 2016-11-29
To fix the bug when testing PDC-1757

bliuredhat:
JIRA: PDC-1757

### DIFF
--- a/pdc/apps/auth/views.py
+++ b/pdc/apps/auth/views.py
@@ -141,16 +141,18 @@ def user_profile(request):
 
 
 def get_users_and_groups(resource_permission):
+    members = dict()
+    inactive_user = {user.username for user in models.User.objects.filter(is_active=False)}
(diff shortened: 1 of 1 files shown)


pdc/apps/auth/views.py
```diff
@@ -141,16 +141,22 @@ def user_profile(request):
 
 
 def get_users_and_groups(resource_permission):
+    group_resource_permission_list = []
     try:
         group_resource_permission_list = get_list_or_404(models.GroupResourcePermission,
                                                          resource_permission=resource_permission)
     except Http404:
```
ycheng-aa: I think here is condition `group_resource_permission_list` being false.
You could use
```
try:
.....
except Http404:
....
else:
...........

```

simozhan:
looks good. seems there are conflicts need to be resolved

PR #434, opened 2017-04-25
Log database errors.

ralphbean:
Really, logging database errors is a good idea.

When this goes wrong, the operator will want to know why!


ralphbean:
Looks like the travis failure is unrelated.

The kobo dep is failing due to the retirement of fedorahosted.org.

PR #436, opened 2017-05-06
Default view of composes should be reversed in order

bliuredhat:
JIRA：PDC-1959


lubomir:
Looks good to me. As it only changes the web UI, there should be no concerns about API stabiilty.

bliuredhat:
+1

PR #453, opened 2017-08-14
Set the settings with settings_test

xiangge:


CONTRIBUTING.markdown
```diff
@@ -35,7 +35,7 @@ merge using fast-forward and push by hand. This will avoid creating a merge
 commit (which makes very little sense for a one-commit pull request).
 
     $ git merge --ff-only feature-branch
-    $ python manage.py test
+    $ python manage.py test --settings=pdc.settings_test
```
hluk: Could this be handled in code instead? I often forgot that there needs to be extra parameter.

xiangge:
@hluk good idea, add the logic in manage.py, please review

bliuredhat:
-1.
Maybe we can add the command in Makefile file,  instead of editing the manage.py !

xiangge:
@bliuredhat yes, you are right, there is already a command "make test" in Makefile.
@hluk and @simozhan, seems we just need to change the "python manage.py test" to "make test" in doc.
What's your thinking?

hluk:
Don't change `manage.py`. I think that file should be handled only by Django itself.

`make test` indeed uses the correct `--settings` argument. The problem is when I want to run specific tests I just use the command `./manage.py test dotted.paths.to.tests` (this breaks few tests).

It's OK to say in documentation that `make test` executes tests.

xiangge:
@hluk  so this is approved?

bliuredhat:
+1; Looks good for me

simozhan:
LGTM

PR #465, opened 2017-09-04
Add signing key in release api

xiangge:
Add signing key in release api

JIRA: PDC-1984

### DIFF
--- a/pdc/apps/release/filters.py
+++ b/pdc/apps/release/filters.py
@@ -39,11 +39,13 @@ class ReleaseFilter(django_filters.FilterSet):
     name = filters.MultiValueFilter(name='name')
     short = filters.MultiValueFilter(name='short')
     version = filters.MultiValueFilter(name='version')
+    sigkey = filters.MultiValueFilter(name='sigkey__key_id')
 
     class Meta:
         model = Release
         fields = ("release_id", "name", "short", "version", 'product_version',
-                  "release_type", "base_product", 'active', 'integrated_with')
+                  "release_type", "base_product", 'active', 'integrated_with',
+                  'sigkey')
 
     def find_has_base_product(self, queryset, value, *args, **kwargs):
         """
--- a/pdc/apps/release/tests.py
+++ b/pdc/apps/release/tests.py
@@ -564,7 +564,8 @@ class ReleaseRESTTestCase(TestCaseWithChangeSetMixin, APITestCase):
         "pdc/apps/release/fixtures/tests/product.json",
         "pdc/apps/release/fixtures/tests/base_product.json",
         "pdc/apps/release/fixtures/tests/product_version.json",
-        "pdc/apps/bindings/fixtures/tests/releasedistgitmapping.json"
+        "pdc/apps/bindings/fixtures/tests/releasedistgitmapping.json",
+        "pdc/apps/common/fixtures/test/sigkey.json"
     ]
 
     def test_create_without_product_version(self):
@@ -574,7 +575,7 @@ class ReleaseRESTTestCase(TestCaseWithChangeSetMixin, APITestCase):
         args.update({"active": True, 'integrated_with': None,
                      'base_product': None, 'product_version': None, 'compose_set': [],
                      'dist_git': None, 'release_id': 'f-20',
-                     'bugzilla': None})
+                     'bugzilla': None, 'sigkey': None})
         self.assertEqual(1, models.Release.objects.filter(release_id='f-20').count())
         self.assertEqual(dict(response.data), args)
         self.assertNumChanges([1])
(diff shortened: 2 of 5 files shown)


pdc/apps/release/filters.py
```diff
@@ -39,6 +39,7 @@ class ReleaseFilter(django_filters.FilterSet):
     name = filters.MultiValueFilter(name='name')
     short = filters.MultiValueFilter(name='short')
     version = filters.MultiValueFilter(name='version')
+    sigkey = filters.MultiValueFilter(name='sigkey__key_id')
```
hluk: Shouldn't "sigkey" be added to `fields` below?

xiangge: I have discussed with @larry-shi , as this "sigkey" is also used by compose, we don't need to put the related release info or keys in "sigkey" table (The api "/sigkey" has already created before), just put this in release's files.

xiangge: > there is an unanswered question from @larry-shi on PDC-1984. Any input on that?

For this sigkey part, I have also discussed with @larry-shi and @simozhan , no need to give a delete operation on the "sigkeys" because the data doesn't increase to much as it just includes the sigkey info. We just keep all the sigkey data here, and give an update operation.

And for release part, I should ask @simozhan for the answer.

xiangge: @hluk Add the "sigkey" in Meta fields, please help to review.

xiangge: @hluk 
But no error without adding this key in fields during my test before.
I just try to remove the fields param to test here, filters still can work, interesting.

xiangge:
@simozhan, I was thinking that as this is just a column of release and all the returns include the "sigkey", so not add special test cases.
But agree with your point, adding more test cases is much better, so add test case on create update and list/query functions. Please review it.

pdc/apps/release/tests.py
```diff
@@ -740,6 +741,29 @@ def test_list_ordered(self):
...
+                "release_type": "ga", "base_product": "product-1", 'sigkey': 'ABCDEF'}
+        response = self.client.post(reverse('release-list'), args)
+        self.assertEqual(status.HTTP_201_CREATED, response.status_code)
+        args.update({'base_product': 'product-1',
+                     'active': True, 'compose_set': [], 'dist_git': None,
+                     'release_id': 'supp-1.1@product-1', 'product_version': None,
+                     'bugzilla': None, 'integrated_with': None, 'sigkey': 'ABCDEF'})
+        self.assertEqual(args, dict(response.data))
+        self.assertNumChanges([1])
+        response = self.client.get(reverse('release-list') + '?base_product=product-1')
```
simozhan: I guess maybe this query for base_product is useless

pdc/apps/release/tests.py
```diff
@@ -1047,7 +1072,7 @@ def setUp(self):
     def test_update(self):
         response = self.client.put(self.url,
                                    {'short': 'product', 'version': '1.0', 'release_type': 'ga',
-                                    'name': 'Our Product'},
+                                    'name': 'Our Product', 'sigkey': 'ABCDEF'},
```
simozhan: Maybe we can have test method test_update_sigkey separately which will verify sigkey update behavior

xiangge: Make sense, add a new function here. please review.

pdc/apps/release/tests.py
xiangge: Yes, removed.

pdc/apps/release/tests.py
```diff
@@ -1275,6 +1298,17 @@ def test_patch_integrated_with(self):
         self.assertEqual(response.data.get('integrated_with'), 'release-2.0')
         self.assertNumChanges([1])
 
+    def test_update_sigkey(self):
+        response = self.client.put(self.url,
+                                   {'short': 'product', 'version': '1.0', 'release_type': 'ga',
+                                    'name': 'Our Product', 'sigkey': 'ABCDEF'},
+                                   format='json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
simozhan: As you update sigkey, why not to verify sigkey's value :)

xiangge: Good point, just change it.

hluk:
It looks OK, but the JIRA mentions "Invite ET guys to Review...". Should anyone in particular review at least some usage examples?

xiangge:
@hluk For the third ac, I think this is more about the data contribution part, and they can use the original "sigkeys" api to add the data in sigkey table. My point is that I think @larry-shi has discussed with ET guys about this requirement,  we can merge the patch into master, then let ET guys review or insert data.

@larry-shi  what's your thinking?

larry-shi:
Do not have to get approval by Errata when try to merge code changes, certainly we should demo this feature with errata guys in stakeholder review meeting or in others meetings.

simozhan:
Ok for the code itself, so approve it.

PR #468, opened 2017-09-15
Add allow_buildroot_push to release API

simozhan:
Add some simple test cases and modify existing cases as well.

JIRA: PDC-1897


hluk:
The JIRA task mentions:

> Only RCM(include above) role has the permission to update this field.

Is it OK not to handle permissions separately in this case?

simozhan:
@hluk Good point. My understanding is only RCM role(except superuser) has permission for release API, so allow_buildroot_push should be part of it, we can see  https://pdc.engineering.redhat.com/rest_api/perms/ , it seems only eng-rcm and superuser has permissions.

PR #471, opened 2017-09-18
Make ComponentBranch API filters case-sensitive

mprahl:
Addresses:
https://pagure.io/releng/issue/6976

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@ settings_local.py
 .idea
 .tox
 pdc.egg-info/
+.vscode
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -6,4 +6,5 @@ settings_local.py
 .idea
 .tox
 pdc.egg-info/
+.vscode
```
hluk: So I'm not the only one who started using vscode for Python :) ...thought the Vim plugin is still quite annoying.

PR #473, opened 2017-09-21
Add allowed_debuginfos for release

xiangge:
Add allowed_debuginfos for release
JIRA： PDC-1971

### DIFF
--- a/pdc/apps/release/serializers.py
+++ b/pdc/apps/release/serializers.py
@@ -15,6 +15,7 @@ from .models import (Product, ProductVersion, Release,
                      validateCPE)
 from . import signals
 from pdc.apps.common.models import SigKey
+from pdc.apps.repository.models import Service
 
 
 class CPEField(serializers.CharField):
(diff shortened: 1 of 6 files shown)


pdc/apps/release/serializers.py
```diff
@@ -91,12 +92,16 @@ class ReleaseSerializer(StrictSerializerMixin, serializers.ModelSerializer):
                                           allow_null=True,
                                           default=None)
     allow_buildroot_push = serializers.BooleanField(default=False)
+    allowed_debuginfos = ChoiceSlugField(slug_field='name',
+                                         many=True, queryset=Service.objects.all(),
+                                         allow_null=True, required=False, default=[])
```
hluk: Shouldn't this be `allow_null=False`?

xiangge: Actually it's not , this value can be set as null ([]).

hluk:
@larry-shi I suggest renaming `allowed_debuginfos` to `debuginfo_services`. Would it be OK or is there any reason for the name?

pdc/apps/release/serializers.py
hluk: `allow_null` just means that you can pass `None` in serializer ([docs](http://www.django-rest-framework.org/api-guide/fields/#allow_null)). Setting value to `null` is disallowed in model (`[]` is not `null`).

Would anything break if you remove it?

xiangge: Aha, thanks for the explain, remove this, and just copy from the ReleaseGroupSerializer.releases, they are similar.

Moved.

larry-shi:
> @larry-shi I suggest renaming allowed_debuginfos to debuginfo_services. Would it be OK or is there any reason for the name?
@hluk @xiangge 
Good idea, how about changing to allowed_debuginfo_services?

hluk:
> @hluk @xiangge
> Good idea, how about changing to allowed_debuginfo_services?

@larry-shi In that case I would need to change `push_targets` to `allowed_push_targets` (in other PR). :) Probably a good idea.

xiangge:
@larry-shi good idea

xiangge:
Changed to allowed_debuginfo_services

simozhan:
Just one thing, next time please commit another patch for your fix if someone already pull the codes and start review, -f push needs the reviewer to resolve conflicts. You can merge the commits together after you get approved by others and then -f push. Thanks :)

xiangge:
@simozhan ok, good suggestions.

xiangge:
@hluk , @simozhan and @larry-shi , pretty cool, thx guys!!

Issue #294, opened 2016-03-03
details about release status


keszybz:
Current information about releases does not provide enough information to distinguish released releases from future and EOL releases.

In the `https://pdc.fedoraproject.org/rest_api/v1/releases/*/` there could be new fields:
- status (one of "development", "released", "eol")
- release_date
- eol_date

The dates would be tentative until the event happens, and could be change over time, and should be set to the real date when the event happens.

This could be used in `dnf-plugin-system-upgrade` and LiveUsb Creator and similar. LUC currently gets the data from https://dl.fedoraproject.org/pub/fedora/linux/releases/ but it's rather slow (has to fetch and parse a lot of small files) and seems brittle.

@lmacken ?

sochotnicky:
The way we generally went about adding things is to add a new resource (i.e. something like "release-schedules" or similar).

Though at least partially - we do support the EOL use case by marking release as "active": "False". Distinguishing devel from releases is not there though...

keszybz:
Normally the whole reason for using an extensible format like XML or json is to be able to freely add new fields ;) I think there'd be a 1:1 mapping between a "release" and "release-schedule" so there isn't much reason to split them. But of course either option would work.

I don't think that the `active` field is useful: in the listings I see EOF releases are not shown (on https://pdc.fedoraproject.org/rest_api/v1/releases/), so active==True for all the ones that are present.

AdamWill:
Now Fedora's pkgdb is being decommissioned, this is an urgent issue for Fedora. The pkgdb 'collections' API listed all releases and had 'under development', 'active', and 'eol' states - so future releases would have 'under development' state, currently-supported stable releases would have 'active' state, and EOL releases would have 'eol' state. I have a project (fedfind) which provides helper functions for discovering 'current' releases based on the pkgdb collections API, and now I'm going to need to switch it to use PDC instead, but this is currently just flat out not possible due to the insufficient detail. For now I've had to cut a fedfind release which just has the current 'correct' answers hard-coded.

I'll also note that the graphical distribution upgrade code for Fedora in gnome-software similarly relies on the pkgdb API to determine what releases are available for upgrade, and it will also need porting to PDC, but it will similarly have trouble until this is fixed, because it obviously needs to distinguish between 'under development' and 'stable' releases.

AdamWill:
@ralphbean @hughsie

hughsie:
[...]

PR #482, opened 2017-10-06
Add the release-schedules endpoint

abompard:
Fixes: #294 #459

About issue #294: to know which releases are currently covered by a specific SLA, you can use the following query (using the SLA `bug_fixes` in this example):
`/rest_api/v1/release-schedules/?sla=bug_fixes&active=1&fields=release`


abompard:
Thanks! I've rebased the commit and fixed the conflict.

simozhan:
Thanks @abompard for pushing this new API. We can not provide the review comments currently, as we think it is not only for Dev's review, it also needs our QE's testing effort to merge this in internal instance. I am assuming this will not affect fedora's instance development. We will return back for merging these features in December 2017. If you want to prioritize the features from fedora, maybe you can discuss this in some meeting.

pypingou:
Looks like there is a conflict on this now.

@simozhan do you know when this PR will be considered now?

abompard:
I can resolve the conflict if necessary, it's probably very simple.

pypingou:
@abompard if you have the time, let's do this so the PR is ready when @simozhan has some time to review it :)

pdc/apps/releaseschedule/serializers.py
```diff
@@ -0,0 +1,57 @@
...
+# Copyright (c) 2017 Red Hat
+# Licensed under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+#
+from rest_framework import serializers
+
+from pdc.apps.release.models import Release
+from pdc.apps.componentbranch.models import SLA
+from pdc.apps.common.fields import ChoiceSlugField
+from pdc.apps.common.serializers import DynamicFieldsSerializerMixin, StrictSerializerMixin
```
lubomir: These mixins should now be imported from `contrib.drf_introspection.serializers`.

pdc/apps/releaseschedule/serializers.py
```diff
@@ -0,0 +1,57 @@
...
+from .models import ReleaseSchedule
+from . import signals
+
+
+class ReleaseScheduleSerializer(DynamicFieldsSerializerMixin,
+                                StrictSerializerMixin,
```
lubomir: Only `StrictSerializerMixin` should be used here, it inherits from `DynamicFieldsSerializerMixin` itself.

abompard:
Thanks for the fixes @lubomir. About the signals, I've seen other components of PDC with signals in these methods, so I figured they are an internal convention and added them too. I can remove them if you prefer.

lubomir:
The signals in `pdc.apps.release` are used by an internal extension. However I would say that turned out as not a very good design decision and I'd rather not repeat the same mistake. Please remove the signals.

abompard:
Done! Thanks.

pdc/apps/releaseschedule/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+    def update(self, instance, validated_data):
+        instance = super(ReleaseScheduleSerializer, self).update(instance, validated_data)
+        if hasattr(instance, 'pk'):
+            # reload to make sure changes in mapping are reflected
+            instance = ReleaseSchedule.objects.get(pk=instance.pk)
+        return instance
+
+    def create(self, validated_data):
+        instance = super(ReleaseScheduleSerializer, self).create(validated_data)
+        return instance
```
lubomir: The `create` and `update` methods can now be dropped.

pdc/apps/releaseschedule/signals.py
```diff
@@ -0,0 +1,28 @@
+#
```
lubomir: This whole file can be removed as well.

PR #489, opened 2017-10-16
Restrict djangorestframework's version

xiangge:
Restrict djangorestframework's version to include more error info.

### DIFF
--- a/pdc.spec
+++ b/pdc.spec
@@ -14,7 +14,7 @@ BuildArch:      noarch
 BuildRequires:  python-setuptools
 BuildRequires:  python-sphinx
 Requires:       Django >= 1.8.1, Django < 1.9.0
-Requires:       django-rest-framework >= 3.1
+Requires:       django-rest-framework >= 3.2
 Requires:       django-rest-framework < 3.3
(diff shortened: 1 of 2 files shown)


hluk:
What "error info" is that?

Are you sure `djangorestframework >= 3.2` is available on supported platforms?

xiangge:
Yes, the version is ok, my local env and qe's env are now 3.2.5.
And also, when you update a data with -d {"aaa": ["1","2"]}, and you put a wrong type data like {“aaa”: "1"}, it will raise "Expected a list of items but got type \"unicode\" in 3.2.5, but in 3.1 there is no such error info, not clear.

hluk:
I meant, whether there is package `django-rest-framework` for all supported platform. But I just noticed that this needs to be defined in [pdc.spec file](https://github.com/product-definition-center/product-definition-center/blob/ad808e85eebe5a667723af9907f33c78fde18993/pdc.spec#L17). Should that be updated too for consistency? And is it possible to use newer version of the package on rhel7 and others?

xiangge:
Oh, yes,I think it's ok for all platforms, I also changed this file, thanks for remind that.

PR #499, opened 2017-10-31
Support OneToOneRel in RelatedNestedOrderingFilter

simozhan:
ReleaseBrewMapping is OneToOneRel to Release, not part of
Release model. Need to be supported seperately.

JIRA: PDC-2208


simozhan:
Will add test cases in internal project.

PR #502, opened 2017-11-01
Add release-files endpoint

xiangge:
Add release-files endpoint

JIRA: PDC-1924


pdc/apps/package/serializers.py
```diff
@@ -221,3 +222,68 @@ class BuildImageRTTTestsSerializer(StrictSerializerMixin, serializers.ModelSeria
...
+    obsolete = serializers.BooleanField(required=False, default=False)
+
+    class Meta:
+        model = models.ReleasedFiles
+        fields = ('id', 'build', 'package', 'file', 'file_primary_key', 'repo',
+                  'released_date', 'release_date', 'created_at', 'updated_at',
+                  'zero_day_release', 'obsolete')
+
+    def validate(self, data):
+        repo_format = data["repo"].content_format
```
hluk: Test `'repo' in data` (and for other keys too) otherwise partial PATCH fails.

pdc/apps/package/views.py
```diff
@@ -532,3 +534,238 @@ def bulk_update(self, *args, **kwargs):
...
+                            "product_id (optional, default=null, nullable)": "int",
+                            "release_id": "string",
+                            "repo_family": "RepoFamily.name",
+                            "service": "Service.name",
+                            "shadow (optional, default=false)": "boolean",
+                            "variant_uid": "string"
+                        },
+                "updated_at (read-only)": "datetime",
+                "zero_day_release (optional, default=false)": "boolean"
+            }
```
hluk: Can you use `%(SERIALIZER)s` instead?

pdc/apps/package/filters.py
```diff
@@ -164,3 +164,26 @@ class BuildImageRTTTestsFilter(django_filters.FilterSet):
...
         model = models.BuildImage
         fields = ('build_nvr', 'test_result', 'image_format')
+
+
+class ReleasedFilesFilter(django_filters.FilterSet):
+    file             = MultiValueFilter()
+    package          = MultiValueFilter()
+    build            = MultiValueFilter()
+    released_date    = MultiValueFilter()
+    release_date     = MultiValueFilter()
```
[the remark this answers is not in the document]
xiangge: As this is Date format, we can just use "2017-11-11" to filter, I have test that, works.

pdc/apps/package/serializers.py
```diff
@@ -221,3 +222,68 @@ class BuildImageRTTTestsSerializer(StrictSerializerMixin, serializers.ModelSeria
     class Meta:
         model = models.BuildImage
         fields = ('id', 'build_nvr', 'format', 'test_result')
+
+
+class RepositoryField(serializers.SlugRelatedField):
```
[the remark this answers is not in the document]
xiangge: There is a name RepoField in repository app.

pdc/apps/package/serializers.py
xiangge: You just can insert/change the repo.id here, the repo info can't change here.

pdc/apps/package/views.py
xiangge: Try this before, it can't print the repo's data. Just a repo id.

pdc/apps/package/filters.py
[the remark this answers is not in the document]
hluk: Yes, but is that useful for anyone to be able to filter by exact date? Also the date format in filter is not verified since it can be any string.

pdc/apps/package/serializers.py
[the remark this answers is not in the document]
hluk: Yes, that's what I meant - use `repository.serializers.RepoField` instead. This is inconsistent with how repositories are serialized in "multi-destination".

pdc/apps/package/serializers.py
hluk: Let me be more clear:
```bash
> pdc_client -r released-files/1 -x PATCH -d '{"repo": 18396}'           
ERROR:pdc_client:503 {"detail":"The server encountered an internal error or misconfiguration and was unable to complete your request."}

> pdc_client -r released-files/1 -x PATCH -d '{"zero_day_release": "true"}'
ERROR:pdc_client:503 {"detail":"The server encountered an internal error or misconfiguration and was unable to complete your request."}
```

pdc/apps/package/views.py
hluk: Again, see how this is done for [multi-destinations](https://github.com/product-definition-center/product-definition-center/blob/a205c467510145206a4e0853412f1b91c8647fa0/pdc/apps/repository/serializers.py#L147).

pdc/apps/common/viewsets.py
```diff
@@ -115,7 +127,11 @@ class ChangeSetDestroyModelMixin(mixins.DestroyModelMixin):
     def perform_destroy(self, obj):
         model_name = get_model_name_from_obj_or_cls(obj)
         obj_id = obj.id
-        obj_content = json.dumps(obj.export())
+        json_item = obj.export()
+        for k in json_item.keys():
+            if isinstance(json_item[k], datetime.date):
+                json_item[k] = str(json_item[k])
```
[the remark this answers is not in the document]
xiangge: Seems not work in export(), this should be get data functions.
I create a  DateToJsonFormat for this. Not sure if it was good to new a class for these few lines before.
And I think put the logic here is okay, because people maybe face the same issue if not add the  conversion here.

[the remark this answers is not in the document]
hluk: Normally the `export()` doesn't take the extra argument (seems to be required only for one API). E.g. from push-targets (much easier to read IMO):
```python
def export(self):
    return {
        "name": self.name,
        "description": self.description,
        "host": self.host,
        "service": self.service.name,
    }
```

pdc/apps/package/filters.py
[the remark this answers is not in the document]
xiangge: What do you mean of exact date? For release and released date, we just let ct insert date type not datetime type.
If you use a wrong string it will show the error info here.
post:
{"release_date":["Date has wrong format. Use one of these formats instead: YYYY[-MM[-DD]]."]}
get:
"detail": [
        "'123213123' value has an invalid date format. It must be in YYYY-MM-DD format."
    ]

pdc/apps/package/views.py
xiangge: I think I know what you mean, but as I mentioned I used "%(SERIALIZER)s" before.
But it just can show the response like the data part:
{
    "file_primary_key": "int",
    "release_date": "date",
    "repo": "Repo.id",
    "package (optional, default=null)": "string",
    "build (optional, default=null)": "string",
    "file (optional, default=null)": "string",
    "created_at (optional, default=null)": "date",
    "released_date (optional, default=null)": "date",
    "updated_at (optional, default=null)": "date",
    "obsolete (optional, default=false)": "boolean",
    "zero_day_release (optional, default=false)": "boolean"
}
This is not what I want, I want to show the repo's detail, so write like this.
I saw your "rest_api/v1/multi-destinations/" codes, same like what I wrote before.
Here the response likes below:
{
    "active (optional, default=true)": "boolean", 
    "destination_repo": "Repo", 
    "global_component": "GlobalComponent.name", 
    "id (read-only)": "int", 
    "origin_repo": "Repo", 
    "subscribers (optional, default=[])": [
        "Person.username"
    ]
}
But I think the Repo part should show all the keys, something like release-components.

xiangge: If I misunderstand your mean, please let me know.

pdc/apps/package/serializers.py
xiangge: Now it's more clear :), and very nice catch.
I have fixed it.

pdc/apps/package/serializers.py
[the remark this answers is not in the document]
xiangge: Why in your RepoField codes, let ct insert "repo": {"id": "18378"} ?

I my new Field class ct just need to insert "repo":  "18378". And I also mention in the __DATA__ part.

pdc/apps/common/viewsets.py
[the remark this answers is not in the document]
xiangge: Hmmm, maybe you are right, put the covert codes here is not good, because almost all the export data don't need to do such check. 
Change this part, just put codes in export function. 
Maybe someone can find what they need to do when face such json error.

pdc/apps/package/filters.py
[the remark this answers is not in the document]
hluk: OK, so even though the filter type is `MultiValueFilter` the format is still required to match the data somehow. Didn't know about this.

Anyway, with current filter user cannot search for records which released after, before or in between some dates. I think that's more useful than filtering by exact date. I would rather see this done in the same way as for the eol field I mentioned above.

pdc/apps/package/serializers.py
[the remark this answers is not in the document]
hluk: > "repo": {"id": "18378"}

This was discussed in PDC-1970, it's used in some other API also.

pdc/apps/package/views.py
hluk: Well, the issue is that after adding a new field in Repo you'll need to update docstrings everywhere. Please don't make me do it, I'll forget to update it anyway. :)

If it's really needed we could change the code which generates the documentation.

xiangge: Yes, you are right, that's a problem, but I think we need to add such repo fields.
I wrote like this because there are other apis with nested fields.
How about to create another task to track this? Then we can do some researching and maybe update all the related codes then?

pdc/apps/package/serializers.py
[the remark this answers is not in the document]
xiangge: How about keep the current RepositoryField?
Here I prefer use "repo": "18378", from my view it is ok :).

pdc/apps/package/filters.py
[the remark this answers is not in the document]
xiangge: oh, sorry for missing the eol part of the first comment, yes, it's a very good idea.
Will add new filters here to get data.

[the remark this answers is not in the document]
xiangge: Commit new codes, please help to review.

pdc/apps/package/views.py
hluk: > How about to create another task to track this? Then we can do some researching and maybe update all the related codes then?

I think, that would take too much time. How about just use `RepoField` type for the "repo" serializer field? :)

pdc/apps/package/serializers.py
```diff
@@ -221,3 +222,79 @@ class BuildImageRTTTestsSerializer(StrictSerializerMixin, serializers.ModelSeria
...
+            repo_format = data["repo"].content_format
+            repo_name = data["repo"].name
+            if str(repo_format) != "rpm":
+                raise serializers.ValidationError(
+                    {'detail': 'Currently we just support rpm type of repo, the type of %s is %s ' % (repo_name, repo_format)})
+
+        if "file_primary_key" in data:
+            d = models.RPM.objects.get(id=data["file_primary_key"])
+            if "build" in data:
+                if data["build"]:
```
hluk: Value of "build" field can be empty?

pdc/apps/package/serializers.py
```diff
@@ -221,3 +222,79 @@ class BuildImageRTTTestsSerializer(StrictSerializerMixin, serializers.ModelSeria
...
+
+            if "file" in data:
+                if data["file"]:
+                    if d.filename != data["file"]:
+                        raise serializers.ValidationError(
+                            {'detail': 'File should be %s' % d.filename})
+                else:
+                    data["file"] = d.filename
+            else:
+                data["file"] = d.filename
```
hluk: So "build", "package" and "file" can contain only specific string depending on RPM data?

If I modify data in rpms, these values here can become incorrect.

Looks like these values don't need to be stored at all since they're available in RPM.

pdc/apps/package/serializers.py
xiangge: Yes, because if in the future there is no field called build in other type of packages.
Currently just use for rpm.

pdc/apps/package/serializers.py
xiangge: Good point, seems there is no way to let it know things have changed.
Will remove the three column there, and give comment on UI to let ct get the three files by file_primary_key.

@simozhan  What's your thinking of this part?

pdc/apps/package/views.py
xiangge: In my view, I still think it is better to keep it here. 
Because the repo has many columns and I add a pdc_endpoint here and also now the build/file/package are also not in get_fields, so keep it here in response is more clear.
But I have changed the data part to  %(WRITABLE_SERIALIZER)s :)

pdc/apps/package/serializers.py
xiangge: I have remove the three columns in db, but list them when ct get data, we can get it from rpms by pk. New patch commit.

pdc/apps/package/views.py
xiangge: I add the patch of #503 here for good looking and test. Will merge these two together when pass the review.

pdc/apps/package/serializers.py
```diff
@@ -221,3 +222,57 @@ class BuildImageRTTTestsSerializer(StrictSerializerMixin, serializers.ModelSeria
...
+class RepositoryField(serializers.SlugRelatedField):
+
+    def __init__(self, **kwargs):
+        super(RepositoryField, self).__init__(slug_field='id', queryset=Repo.objects.all(), **kwargs)
+
+    def to_representation(self, value):
+        export_data = value.export()
+        export_data["id"] = value.id
+        export_data["content_format"] = {"name": value.content_format.name,
+                                         "pdc_endpoint": value.content_format.pdc_endpoint}
```
hluk: This field should be visible in top level of the new endpoint instead.

pdc/apps/repository/models.py
```diff
@@ -37,6 +37,7 @@ def export(self):
 class ContentFormat(models.Model):
     # rpm, kickstart, iso
     name          = models.CharField(max_length=50, unique=True)
+    pdc_endpoint = models.CharField(max_length=200, null=True)
```
hluk: Updating this field can invalidate some `file_primary_key` values.

pdc/apps/package/serializers.py
```diff
@@ -221,3 +222,57 @@ class BuildImageRTTTestsSerializer(StrictSerializerMixin, serializers.ModelSeria
...
+
+        # if there are other types support, should add codes check if the primary key in right table
+
+        return data
+
+    def to_representation(self, value):
+        rep_data = super(ReleasedFilesSerializer, self).to_representation(value)
+        if "file_primary_key" in rep_data and "repo" in rep_data:
+            if rep_data["repo"]["content_format"]["name"] == "rpm":
+                d = models.RPM.objects.get(id=rep_data["file_primary_key"])
```
hluk: This can throw `models.RPM.DoesNotExist`.

pdc/apps/repository/models.py
xiangge: Yes, this value can be changed in admin page. not from UI or rest api.
And I think this could be set once. 
And if it's changed, all the values will changed in release_files-> repo->content-format->pdc_endpoint, but the content format type can't be changed, as it is in repo.
It's not a problem, so what do your mean?

pdc/apps/package/models.py
```diff
@@ -357,6 +359,37 @@ def export(self, fields=None):
...
+    def export(self, fields=None):
+        _fields = (set(['file_primary_key', 'repo',
+                        'released_date', 'release_date', 'created_at',
+                        'updated_at', 'zero_day_release', 'obsolete'])
+                   if fields is None else set(fields))
+        result = model_to_dict(self, fields=_fields)
+        for k in ["release_date", "released_date", "created_at", "updated_at"]:
+            if k in result:
+                result[k] = str(result[k])
+        return result
```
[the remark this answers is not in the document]
xiangge: Yes, but you can see RPM and BuildImage are also do like this.
And the difference between these ones and others is that there is a "fields=None", this can let you choose the fields you want to show. 
I think this is better way, agree?

pdc/apps/package/serializers.py
xiangge: Which level you mean?
Currently there is a content_format in repo, so we don't need add a content format again in release-files, just get the repo means get the content_format, and also if we move the pdc_endpoint out in repo, this is also not good because if you have looked at the repo's api you can see there is just content_format.name here, so I think keep it here is more clearly, and that's why I put a dict in the docstring.

xiangge: # This will make APIs inconsistent. Maybe we can improve RepoField to take simple ID value in addition to dict.
For me, I think we'd better just keep it here, sync the two classes will break other api. 
We can create an new issue to do this.

pdc/apps/package/serializers.py
xiangge: Do you mean these error? It can return to the response.
{"detail":"Not found:  RPM matching query does not exist."}

pdc/apps/package/serializers.py
hluk: This changes the way `content_format` is serialized - it can break some client code which expects the type to be simple string. Also, I think it's more difficult for user to access the `pdc_endpoint` value. So I propose to serialize it with released-files serializer directly and have it on the top level as `repo` field.

pdc/apps/package/models.py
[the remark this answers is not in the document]
hluk: Value of `fields` is not used anywhere. I don't like this is duplicated everywhere without being ever used. It makes the code unnecessarily complicated.

pdc/apps/package/serializers.py
hluk: OK, looks good.

pdc/apps/repository/models.py
hluk: Seems the purpose of `pdc_endpoint` field only to tell user where to find the data for `file_primary_key`. It's not used in code. think it's better not to store `pdc_endpoint` in DB, just serialize it as an derived value - code already knows which endpoint to use for `file_primary_key`. Is that correct?

pdc/apps/package/models.py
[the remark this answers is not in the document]
xiangge: When I init a field class in another serializer just list primary_key, released_data, that would be used. Just think this implement is good, so kept that part before.

pdc/apps/package/serializers.py
xiangge: That makes sense, I have moved it to the top.

pdc/apps/repository/models.py
xiangge: I assume it's not, even the first people who inserted the data know the endpoint, but maybe the other guys get the data not know the endpoint.
We keep the endpoint here, that can make sure the api uris are correct when ct is using.

BTW: Have moved the pdc_endpoint to the top level.

pdc/apps/package/models.py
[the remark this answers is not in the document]
hluk: Usually I'm against the idea adding a code that could be potentially used later, because it won't be used in most cases :). If you think it's OK in this case, let's keep it.

pdc/apps/package/views.py
```diff
@@ -590,7 +588,8 @@ def list(self, *args, **kwargs):
                 "zero_day_release (optional, default=false)": "boolean",
                 "build (read-only)": "string",
                 "created_at (read-only)": "datetime",
-                "file (read-only)": "string"
+                "file (read-only)": "string",
+                "pdc_endpoint": "string"
```
hluk: Shouldn't this be read-only field?

xiangge: Yes, correct, changed that, thanks :)

pdc/apps/package/models.py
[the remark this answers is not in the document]
xiangge: Yes, you are right, currently it has no use, but let's keep it, maybe we will use it :)

pdc/apps/repository/models.py
hluk: Currently `pdc_endpoint` for all content types is null. Updating the value will require new migrations or manual action to set the values. This is unnecessary complication since the code already knows which data model to use for which content type. I.e. it's better not to store `pdc_endpoint` in DB.

xiangge: For this column, I have discussed with Ren and asked for the reasons before, actually he thinks this needs to store in db :).
I think adding it here is for that people can get the official endpoint name from pdc, from Ren's view, it's an important thing I think :)
And it's easy to add the value from admin although currently we just need the rpms.

pdc/apps/package/serializers.py
```diff
@@ -221,3 +222,60 @@ class BuildImageRTTTestsSerializer(StrictSerializerMixin, serializers.ModelSeria
...
+
+        # if there are other types support, should add codes check if the primary key in right table
+
+        return data
+
+    def to_representation(self, value):
+        rep_data = super(ReleasedFilesSerializer, self).to_representation(value)
+        if "repo" in rep_data and "pdc_endpoint" in rep_data["repo"]:
+            rep_data["pdc_endpoint"] = rep_data["repo"]["pdc_endpoint"]
+            rep_data["repo"].pop("pdc_endpoint")
```
hluk: Looks like it would be much easier to just set `rep_data["pdc_endpoint"] = "rpm"` below and remove the `pdc_endpoint` from data model.

xiangge: This is a hard code, maybe one day we change that, we don't want to change the codes.
I am thinking that as there are multi types here, we should do an automatically way to do this, but not just add the if/else.

hluk: Why would we want to change that? That would break API. Currently, we need to change the code anyway to support new content type.

pdc/apps/package/models.py
```diff
@@ -357,6 +359,37 @@ def export(self, fields=None):
         return result
 
 
+class ReleasedFiles(models.Model):
+    file_primary_key = models.IntegerField(blank=False, default=0)
```
hluk: Looks like Django supports foreign keys to multiple models [1].

Did you try it?

BTW, why is there `default=0`?

xiangge: You mean contenttypes?
I have looked at it for sometime.
Here the contenttypes is about all the tables,  I am doing the research on that

default=0:
As this is not a foreign key, so when do the makemigrations it need a default value.

pdc/apps/package/serializers.py
xiangge: I just think anything could happen :), also when we design a new endpoint with new type, we can't make sure if we never change that.

xiangge:
@hluk 
> Use RepoField instead RepositoryField (inconsistent API, confusing names)
I have asked Chuang's advise about this, he thinks we should keep the repo input the same when this is used by other places. This makes sense to me and as you mentioned ct wants to insert {"id": int} in multi destinations, so I can just keep the same. 
And also after discuss with Chuang Yaobin, we decide to remove the pdc_endpoint in release-files, and just add a docstring here to let ct choose if they want to get the endpoint name of not.

> Remove data examples from docstrings (impossible to maintain)
I find a good way, just need to edit RepoField's doc_format.

For other item, we can resolve later:
New pdc_endpoint field (maybe shouldn't be stored in DB, or used at all)
 
Patch commit, please review.

PR #503, opened 2017-11-01
Add pdc_endpoint in content format table

xiangge:
Add pdc_endpoint in content format table

JIRA: PDC-2362

### DIFF
--- a/pdc/apps/repository/migrations/0005_contentformat_pdc_endpoint.py
+++ b/pdc/apps/repository/migrations/0005_contentformat_pdc_endpoint.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('repository', '0004_multidestination'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='contentformat',
+            name='pdc_endpoint',
+            field=models.CharField(max_length=200, null=True),
+        ),
+    ]
--- a/pdc/apps/repository/models.py
+++ b/pdc/apps/repository/models.py
@@ -37,6 +37,7 @@ class Service(models.Model):
 class ContentFormat(models.Model):
     # rpm, kickstart, iso
     name          = models.CharField(max_length=50, unique=True)
+    pdc_endpoint  = models.CharField(max_length=200, null=True)
     description   = models.CharField(max_length=200)
 
     def __unicode__(self):
--- a/pdc/apps/repository/serializers.py
+++ b/pdc/apps/repository/serializers.py
@@ -120,7 +120,7 @@ class ContentFormatSerializer(StrictSerializerMixin, serializers.ModelSerializer
 
     class Meta:
         model = models.ContentFormat
-        fields = ('name', 'description',)
+        fields = ('name', 'pdc_endpoint', 'description',)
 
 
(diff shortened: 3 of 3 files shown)


xiangge:
Base on Ren Yang's solution.
Endpoint 2. deliverable_content_type
```
{
id,
name:
pdc_endpoint.
}
```
This new endpoint is dup with the content format, so just need add a pdc_endpoint here.

hluk:
> This new endpoint is dup with the content format, so just need add a pdc_endpoint here.

OK, but this should be used in PDC-1924, right?

Shouldn't "name" be foreign key to "content-delivery-content-formats"?

xiangge:
`OK, but this should be used in PDC-1924, right?
`
Yes.
`Shouldn't "name" be foreign key to "content-delivery-content-formats"?`
In content-delivery-repo's api, it shows the name, so we can get it there, and it is unique in content-delivery-content-formats. So people can get content-delivery-content-formats' data by name.
I think this pdc_endpoint is designed for future, currently I didn't see the use case.

hluk:
If the new API is not used why implement it now? I think it should be implemented and used in #502.

I still don't understand why to use string instead of foreign key to content format. Using the key makes the relation stronger and clearer.

xiangge:
Here is one data in new release-files:
        ```
   {
            "id": 33,
            "build": null,
            "package": null,
            "file": null,
            "file_primary_key": 2,
            "repo": {
                "release_id": "rhel-6.7",
                "content_category": "binary",
                "shadow": false,
                "repo_family": "beta",
                "id": 19348,
                "name": "rhel-6-desktop-beta-isos__6Client__i386",
                "variant_uid": "Client",
                "product_id": 68,
                "service": "pulp",
                "arch": "i386",
                "content_format": "iso"
            },
            "released_date": null,
            "release_date": "2017-11-11",
            "created_at": "2017-10-31T03:26:27.220497Z",
            "updated_at": "2017-10-31T03:27:47.941642Z",
            "zero_day_release": false,
            "obsolete": false
        }
```

You can see there is a content_format in repo, the foreign key is content_format's id in repo.
So we don't need to add the content_format's foreign key here. 
Add this pdc_endpoint is more likely to store the endpoints name of each type in future.

hluk:
The problem is that this API is not used anywhere - i.e. no need to keep it. I would expect the other PR to use `pdc_endpoint` to find correct data model.

xiangge:
Yes, that makes sense, will close this pr.

xiangge:
@hluk  I have discussed with Ren again for this part, then I think we should add this column here, we can list the pdc_endpoint in release_files data, then ct can get the endpoint name and do further things.

Like:

>        {
>             "id": 39,
>             "file_primary_key": 5,
>             "repo": {
>                 "release_id": "rhel-6.8",
>                 "content_category": "binary",
>                 "shadow": false,
>                 "repo_family": "beta",
>                 "id": 18378,
>                 "name": "rhel-6-desktop-beta-rpms__6Client__i386",
>                 "variant_uid": "Client",
>                 "product_id": 68,
>                 "service": "pulp",
>                 "arch": "i386",
>                 "content_format": {
>                     "name": "rpm",
>                     "pdc_endpoint": "rpms"
>                 }
>             },
>             "released_date": null,
>             "release_date": "2017-11-12",
>             "created_at": "2017-10-31T08:59:38.007696Z",
>             "updated_at": "2017-10-31T08:59:46.888620Z",
>             "zero_day_release": false,
>             "obsolete": false
>         }

xiangge:
Close this pr, as I have added the codes in #502

PR #516, opened 2018-01-05
Add the "variant_context" field to the unreleasedvariants API

mprahl:
This PR adds the "variant_context" field to the unreleasedvariants API. This field is used to identify a module build with the same name (variant_name), stream (variant_version), and version (variant_release).

This PR also fixes a bug where uniqueness is not enforced on the "variant_uid" field.

### DIFF
--- a/pdc/apps/module/filters.py
+++ b/pdc/apps/module/filters.py
@@ -17,6 +17,7 @@ class UnreleasedVariantFilter(django_filters.FilterSet):
     variant_type        = django_filters.CharFilter(name='variant_type', lookup_type='iexact')
     variant_version     = django_filters.CharFilter(name='variant_version', lookup_type='iexact')
     variant_release     = django_filters.CharFilter(name='variant_release', lookup_type='iexact')
+    variant_context     = django_filters.CharFilter(name='variant_context', lookup_type='iexact')
     active              = CaseInsensitiveBooleanFilter()
     koji_tag            = django_filters.CharFilter(name='koji_tag', lookup_type='iexact')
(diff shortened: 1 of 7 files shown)


lubomir:
What does the context mean? When other tooling is updated to use this, will `variant_uid` change to include the context as well? (As far as I can tell PDC does not enforce any particular value for the field.)

Code wise the change looks good to me.

In the long term we should look into cleaning up the naming. Current field names in this endpoint are really confusing. 🤔

mprahl:
@lubomir the context will be used as a way to identify what module dependencies the module was built with and what modules it requires at run-time. This is because we plan to have multiple module builds per modulemd in the future (module stream expansion), with the only difference being the module's module dependencies. This is a way to make each build unique.

I agree that we should rename these fields as they are confusing. In addition, I'd like to make `variant_uid` be a read-only attribute that combines `name-stream-version-context`, but that will be a separate PR.

mprahl:
@lubomir Thank you for the review. Is this good to merge?

lubomir:
Making the uid unique breaks backward compatibility, but it is fixing a bug, so it should be okay. (If there were two modules with the same UID, it would break detail view in some way, as the UID is used in URL there.)

What do you think, @hluk?

hluk:
> Making the uid unique breaks backward compatibility, but it is fixing a bug, so it should be okay. (If there were two modules with the same UID, it would break detail view in some way, as the UID is used in URL there.)

I think it could be an issue if there are more entries with same UID in DB - Django might complain and this may need to be resolved manually.

For me it's good to merge (it also fixes PDC-2097).

PR #518, opened 2018-01-19
Proof of concept on immutable identifiers

lubomir:
The tests are failing because of the added primary key field.

It should work for releases, release variants and compose trees.


contrib/bulk_operations/bulk_operations.py
```diff
@@ -190,12 +190,16 @@ def register(self, prefix, viewset, base_name=None):
     def get_lookup_regex(self, viewset, lookup_prefix=''):
         """
         For viewsets using the MultiLookupFieldMixin, it is necessary to
-        construct the lookup_value_regex attribute here.
+        construct the lookup_value_regex attribute here so that it allows using
+        either the composite value or databse primary key.
```
hluk: s/databse/database/

pdc/apps/common/tests.py
```diff
@@ -785,3 +785,40 @@ def test_delete_notification(self, mock_router):
...
+                self.get_queryset_called = True
+                return self.queryset
+
+        class DummyView(viewsets.NumericIDMixin, Base):
+            lookup_field = 'short'
+
+        self.View = DummyView
+
+    def test_retrieve_by_id(self):
+        view = self.View(short=1)
```
hluk: Not sure why is this needed. Why not use everywhere `pk` with numeric value instead of `short`?

pdc/apps/release/serializers.py
```diff
@@ -64,7 +64,7 @@ class ProductSerializer(StrictSerializerMixin, serializers.ModelSerializer):
 
     class Meta:
         model = Product
-        fields = ('name', 'short', 'active', 'product_versions', 'allowed_push_targets')
+        fields = ('name', 'short', 'active', 'product_versions', 'allowed_push_targets', 'id')
```
hluk: Would it be OK to use name `pk` instead of `id`? Endpoint "release-variants" has `id` field which is string.

pdc/apps/common/tests.py
lubomir: In the tests this is basically directly translated into `self.kwargs`. In real viewsets DRF is extracting the values from url regex, and the field name is determined by the name in the regex, which always corresponds to `lookup_field`. For the simple cases of one field identifier this is always the name of the human readable id field. For example the regex for products would be `(?P<short>[^/.]+)`. The field will match either number or string.

For multi field views the regex is more complicated, for example release variants will have  `(?P<pk>\d+)|(?P<release__release_id>[^/]+)(?P<variant_uid>[^/]+)`, so `self.kwargs` will either have `pk` key or the two string identifiers.

hluk: The regex from `NumericIDMixin` contains `?<pk>` so why doesn't `self.View(pk=1)` work? It's quite confusing that `short` gets translated to `pk`.

pdc/apps/release/tests.py
```diff
@@ -126,7 +126,7 @@ def test_create(self):
         args = {"name": "Fedora", "short": "f"}
         response = self.client.post(reverse('product-list'), args)
         self.assertEqual(status.HTTP_201_CREATED, response.status_code)
-        args.update({'active': False, 'product_versions': [], 'allowed_push_targets': []})
+        args.update({'active': False, 'product_versions': [], 'allowed_push_targets': [], "pk": 3})
```
hluk: Single quotes.

pdc/apps/release/tests.py
```diff
@@ -160,7 +160,15 @@ def test_get(self):
...
-                          "allowed_push_targets": []})
+                          "allowed_push_targets": [], 'pk': 1})
+
+    def test_get_by_pk(self):
+        response = self.client.get(reverse('product-detail', args=[1]))
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
+        self.assertEqual(dict(response.data),
+                         {"short": "product", "name": "Test Product",
+                          "product_versions": [], "active": False,
+                          "allowed_push_targets": [], 'pk': 1})
```
hluk: Double quotes?

pdc/apps/common/tests.py
lubomir: It's actually also wrong. It does not work correctly with bulk operations on multi field views.

lubomir:
I tried to clarify what the test does. Is it less confusing now?

hluk:
It's clear now, thanks. Does someone else need to approve this? It's a bigger change.

lubomir:
I added more people for review, that's a fair point. Since we don't want this in the next release, I think it's ok if it stays on review for a while.

PR #519, opened 2018-01-23
Container build throws 404 error on graphviz repo

kellinm:
- external graphviz repository URL is no longer valid
- drop the dependency on the external graphviz repository since
  a previous commit changed the container to use Fedora which
  provides a suitable version of graphviz

Signed-off-by: Robert Marshall <rmarshall@redhat.com>


lubomir:
The image still builds without the repo. 👍

PR #520, opened 2018-01-24
Drop contact browser from examples

lubomir:
It's not really updated and some of listed dependencies have known security vulnerabilities. Additionally, if someone wants to create a client UI, they should probably not take this as inspiration...


lubomir:
I don't see any point in keeping this example. This same thing is available in a separate repo. I'll try to update the partners app.

PR #521, opened 2018-01-24
Add v2 of the "unreleasedvariants" API called "module"

mprahl:
Add v2 of the "unreleasedvariants" API called "module" and correct the docs in "unreleasedvariants"

To use v1 of the API, continue using the "unreleasedvariants" API but for v2, use the "module" API.

The new version of the API does the following:
* Renames "variant_name" to "name"
* Renames "variant_version" to "stream"
* Renames "variant_release" to "version"
* Renames "variant_context" to "context"
* Renames "variant_uid" to "uid"
* Drops the usage of "variant_type" (now set to "module" in the DB)
* Drops the usage of "variant_id" (now set to the module name in the DB)
* "uid" is now read-only and the value is derived as "name:stream:version:context"

### DIFF
--- a/pdc/apps/module/__init__.py
+++ b/pdc/apps/module/__init__.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2015 Red Hat
+# Copyright (c) 2018 Red Hat
 # Licensed under The MIT License (MIT)
 # http://opensource.org/licenses/MIT
 #
--- a/pdc/apps/module/models.py
+++ b/pdc/apps/module/models.py
@@ -6,61 +6,49 @@ from django.db import models
 from pdc.apps.package.models import RPM
 
 
-class UnreleasedVariant(models.Model):
-    # Not the variant from compose ... which back references compose
-    variant_id          = models.CharField(max_length=100, blank=False)
-    variant_uid         = models.CharField(max_length=200, blank=False, unique=True)
-    variant_name        = models.CharField(max_length=300, blank=False)
-    variant_type        = models.CharField(max_length=100, blank=False)
-    # variant_version/_release are _not_ distribution versions/releases
-    variant_version     = models.CharField(max_length=100, blank=False)
-    variant_release     = models.CharField(max_length=100, blank=False)
-    # Default to '00000000' for now since this field will only be used once
-    # other tooling is updated to supply this value. Eventually, this should
-    # not have a default.
-    variant_context     = models.CharField(max_length=100, blank=False, default='00000000')
-    active              = models.BooleanField(default=False)
-    koji_tag            = models.CharField(max_length=300, blank=False)
-    modulemd            = models.TextField(blank=False)
-    rpms                = models.ManyToManyField(RPM)
+class Module(models.Model):
+    variant_id  = models.CharField(max_length=100, blank=False)
+    uid         = models.CharField(max_length=610, blank=False, unique=True)
+    name        = models.CharField(max_length=300, blank=False)
+    stream      = models.CharField(max_length=100, blank=False)
+    version     = models.CharField(max_length=100, blank=False)
+    context     = models.CharField(max_length=100, blank=False)
+    # No new instances should have a `type` value other than `module`. This is here just for
+    # backwards compatibility with the `unreleasedvariants` API.
+    type        = models.CharField(max_length=100, blank=False, default='module')
+    active      = models.BooleanField(default=False)
+    koji_tag    = models.CharField(max_length=300, blank=False)
+    modulemd    = models.TextField(blank=False)
+    rpms        = models.ManyToManyField(RPM)
 
     class Meta:
-        ordering = ("variant_name", "variant_version", "variant_release",
-                    "variant_context")
+        ordering = ('name', 'stream', 'version', 'context')
         unique_together = (
-            ("variant_name", "variant_version", "variant_release",
-             "variant_context"),
+            ('name', 'stream', 'version', 'context'),
         )
 
     def __unicode__(self):
-        return u"%s" % (self.variant_uid, )
+        return u'%s' % (self.uid, )
 
     def export(self):
-        result = {
+        return {
             'variant_id': self.variant_id,
-            'variant_uid': self.variant_uid,
-            'variant_name': self.variant_name,
-            'variant_type': self.variant_type,
-            'variant_version': self.variant_version,
-            'variant_release': self.variant_release,
-            'variant_context': self.variant_context,
+            'uid': self.uid,
+            'name': self.name,
+            'stream': self.stream,
+            'version': self.version,
+            'context': self.context,
+            'type': self.type,
             'active': self.active,
             'koji_tag': self.koji_tag,
             'modulemd': self.modulemd,
+            'rpms': [obj.export() for obj in self.rpms.all()],
             'runtime_deps': [v.dependency for v in self.runtime_deps.all()],
             'build_deps': [v.dependency for v in self.build_deps.all()],
         }
 
-        result["rpms"] = []
-        objects = self.rpms.all()
-        for obj in objects:
-            result["rpms"].append(obj.export())
 
-        return result
-
-
-class VariantDependency(models.Model):
-
+class ModuleDependency(models.Model):
     dependency = models.CharField(max_length=300)
     stream = models.CharField(max_length=300)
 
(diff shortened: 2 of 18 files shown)


pdc/apps/unreleasedvariant/tests.py
```diff
@@ -0,0 +1,331 @@
+# -*- coding: utf-8 -*-
```
mprahl: This is the same exact `tests.py` that was at `pdc/apps/module/tests.py`. The only difference is the last test function at the bottom.

pdc/apps/unreleasedvariant/tests.py
```diff
@@ -0,0 +1,331 @@
...
+            'variant_release': "1", 'variant_type': 'module',
+            'variant_context': 'a23d56a8', 'koji_tag': "module-core-0-1",
+            'modulemd': 'foobar'
+        }
+        response = self.client.post(url, data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        uv = Module.objects.filter(uid='coretest123').first()
+        self.assertEqual(uv.context, 'a23d56a8')
+
+    def test_v2_compatibility_unreleasedvariant(self):
```
mprahl: This is the only test I added to this file. The rest came from the original `pdc/apps/module/tests.py`.

pdc/apps/unreleasedvariant/views.py
```diff
@@ -0,0 +1,205 @@
...
+                        "modulemd": string,
+                        "active": bool,
+                        "runtime_deps": [{"dependency": string, "stream": string}, ... ],
+                        "build_deps": [{"dependency": string, "stream": string}, ... ],
+                        "rpms": [string, ... ]
+                    },
+                    ...
+                ]
+            }
+        """
```
hluk: You can remove these docstrings if the generated API documentation looks OK.

It would be difficult to keep these up-to-date in future (nobody updates comments).

pdc/apps/module/routers.py
```diff
@@ -1,5 +1,10 @@
...
+# Copyright (c) 2018 Red Hat
+# Licensed under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+#
+from pdc.apps.module.views import ModuleViewSet
 from pdc.apps.utils.SortedRouter import router
 
 
-router.register(r'unreleasedvariants', views.UnreleasedVariantViewSet)
+router.register(r'module', ModuleViewSet, base_name='module')
```
lubomir: A bit of bikeshedding: all REST-like urls use plural, so this should as well probably.

pdc/apps/module/models.py
```diff
@@ -1,80 +1,62 @@
...
-    modulemd            = models.TextField(blank=False)
-    rpms                = models.ManyToManyField(RPM)
+class Module(models.Model):
+    variant_id  = models.CharField(max_length=100, blank=False)
+    uid         = models.CharField(max_length=200, blank=False, unique=True)
+    name        = models.CharField(max_length=300, blank=False)
+    stream      = models.CharField(max_length=100, blank=False)
+    version     = models.CharField(max_length=100, blank=False)
+    context     = models.CharField(max_length=100, blank=False)
+    type        = models.CharField(max_length=100, blank=False, default='module')
```
lubomir: ~~Out of curiosity: what other types can there be for a module?~~ Assuming this is for backwards compatibility with the v1 API (as the comment there suggests), it might be good to mention it here that no new instances should have a type different than *module*.

pdc/apps/unreleasedvariant/admin.py
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright (c) 2018 Red Hat
+# Licensed under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+#
+from pdc.apps.common.register_to_admin import register
+
+register('unreleasedvariant')
```
lubomir: If there are no models in the app, then registering to admin does not make much sense to me. Can this file be removed?

pdc/apps/unreleasedvariant/apps.py
```diff
@@ -0,0 +1,14 @@
...
+#
+from django.apps import AppConfig
+
+
+class UnreleasedVariantConfig(AppConfig):
+    name = 'pdc.apps.unreleasedvariant'
+
+    def ready(self):
+        from pdc.apps.utils.utils import connect_app_models_pre_save_signal
+        connect_app_models_pre_save_signal(self)
```
lubomir: Again, since there are no models, there should also be no signals.

pdc/apps/unreleasedvariant/views.py
```diff
@@ -0,0 +1,205 @@
...
+from pdc.apps.unreleasedvariant.serializers import UnreleasedVariantSerializer
+from pdc.apps.unreleasedvariant.filters import UnreleasedVariantFilter
+
+
+class UnreleasedVariantViewSet(viewsets.PDCModelViewSet):
+    """
+    ##Overview##
+
+    This page shows the usage of the **Module API**, please see the following for more details.
+    """
```
lubomir: It would be nice to mention here that a new version is available should be preferred in new applications.

pdc/apps/unreleasedvariant/views.py
hluk: Sorry, just realized that it's still not possible to remove these docstrings. Would be at least nice to use `%(SERIALIZER)s` where possible.

pdc/apps/module/models.py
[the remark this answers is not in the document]
mprahl: In this situation, overshadowing the built-in `type` function is okay because it's only overshadowed in the class variables. In all other parts of the class, it has to be referred by `self.type`. By keeping it as `type`, it makes it more intuitive everywhere else in the code, especially in queries.

mprahl: Good point. Will add a comment here.

pdc/apps/module/__init__.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2015 Red Hat
+# Copyright (c) 2018 Red Hat
```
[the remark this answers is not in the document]
mprahl: I'll make these changes in a separate commit. I made the change in this PR because I felt like the year should be updated since this basically felt like a rewrite of the `module` Django app. If you disagree, I can get rid of the commit.

pdc/apps/module/routers.py
mprahl: Good point. I'll change this.

pdc/apps/unreleasedvariant/admin.py
mprahl: Sure

pdc/apps/unreleasedvariant/apps.py
mprahl: I'll remove it

pdc/apps/unreleasedvariant/views.py
mprahl: Good point. I forgot to put that.

pdc/apps/unreleasedvariant/views.py
mprahl: Good idea. I had forgotten PDC could do this.

mprahl:
@hluk and @lubomir thank you for the reviews! I've addressed your comments and please review again when you can.

pdc/apps/module/models.py
```diff
@@ -1,80 +1,64 @@
...
-    # other tooling is updated to supply this value. Eventually, this should
-    # not have a default.
-    variant_context     = models.CharField(max_length=100, blank=False, default='00000000')
-    active              = models.BooleanField(default=False)
-    koji_tag            = models.CharField(max_length=300, blank=False)
-    modulemd            = models.TextField(blank=False)
-    rpms                = models.ManyToManyField(RPM)
+class Module(models.Model):
+    variant_id  = models.CharField(max_length=100, blank=False)
+    uid         = models.CharField(max_length=200, blank=False, unique=True)
```
hluk: Since value for `uid` is concatenation of the "unique-together" fields below, `max_length=200` is insufficient.

mprahl: Good catch! This was fixed in the latest rebase.

mprahl:
@hluk and @lubomir Thank you for the review. This is ready for review again.

mprahl:
Thanks for the approvals! Is this good to merge now?

PR #528, opened 2018-02-02
Fix displaying deleted composes

lubomir:
When a compose is marked as deleted, it should not show up by default in some places. The compose list has a filter, so there it's ok.

This patch fixes and tests that compose_set in /releases/ API does not include deleted composes (either directly connected or linked).

Also the lower level method used to get latest compose for a release is tested to not include deleted ones. This should fix RPM Mapping API end-point.

The APIs to get compose by release or product version and RPM NVR are updated to not return deleted composes.

JIRA: PDC-2533

### DIFF
--- a/pdc/apps/compose/tests.py
+++ b/pdc/apps/compose/tests.py
@@ -25,6 +25,13 @@ import pdc.apps.common.models as common_models
 from . import models
 
 
+def _delete_compose(compose_id):
+    """Helper function to mark a compose as deleted."""
+    c = models.Compose.objects.get(compose_id=compose_id)
+    c.deleted = True
+    c.save()
+
+
 class ComposeModelTestCase(TestCase):
     fixtures = [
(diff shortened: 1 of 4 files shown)


hluk:
Should deleted composes be hidden everywhere except /composes/ end-point?

In which situations should be existing compose deleted? Also, looking at API docs, it seems it's possible to create "deleted" compose.

hluk:
> Also, looking at API docs, it seems it's possible to create "deleted" compose.

Oops, just noticed that there is no create API for composes.

lubomir:
> Should deleted composes be hidden everywhere except /composes/ end-point?

I would say yes. All places that work with composes should only consider non-deleted ones. In the composes end-point the status is displayed, so it can have them.

The use case for deleting composes is to have a way to remove incorrect data (for example caused by configuration error) from messing up rpm mapping for a release without having to wait for a correct compose.

PR #531, opened 2018-02-05
Include radix in parseInt call

lubomir:
The default is not guaranteed to be 10.

### DIFF
--- a/pdc/static/js/datatable_active_filter.js
+++ b/pdc/static/js/datatable_active_filter.js
@@ -2,7 +2,7 @@
     "use strict";
     $(document).ready(function() {
         if ($('select').is('#active_filter')){
-            var page_size = parseInt($('select').attr("value"));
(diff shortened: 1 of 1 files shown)


hluk:
The default is what?! It would be evil from implementers not to use 10 as default. :)

lubomir:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt


>   * If the input string begins with "0x" or "0X", radix is 16 (hexadecimal) and the remainder of the string is parsed.
>   * If the input string begins with "0", radix is eight (octal) or 10 (decimal).  Exactly which radix is chosen is implementation-dependent.  ECMAScript 5 specifies that 10 (decimal) is used, but not all browsers support this yet.  For this reason always specify a radix when using parseInt.
>   * If the input string begins with any other value, the radix is 10 (decimal).

hluk:
Oh, OK, that makes sense.

PR #532, opened 2018-02-05
Fix access to modules with dot in UID

lubomir:
The default router will not allow the field in URL to contain a dot: it fails to create the URL in that case. This patch sets a regex so that the UID can be anything without a slash.

JIRA: PDC-2535

### DIFF
--- a/pdc/apps/common/viewsets.py
+++ b/pdc/apps/common/viewsets.py
@@ -248,7 +248,11 @@ class PDCModelViewSet(StrictQueryParamMixin,
     PDC common ModelViewSet.
     With `StrictQueryParam`, `ProtectOnDelete` and `ChangeSetModel`
     """
-    pass
+
+    # The field used in URL can not contain a . character by default. This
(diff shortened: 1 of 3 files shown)


pdc/apps/unreleasedvariant/views.py
```diff
@@ -21,6 +21,7 @@ class UnreleasedVariantViewSet(viewsets.PDCModelViewSet):
     filter_class = UnreleasedVariantFilter
     serializer_class = UnreleasedVariantSerializer
     lookup_field = 'uid'
+    lookup_value_regex = r'[^/]+'
```
hluk: Can this be moved to some base class? And ideally use constant with descriptive name or add comment so this is not accidentally removed again.

ralphbean:
\ó/

ralphbean:
Attempting to build -> https://koji.fedoraproject.org/koji/taskinfo?taskID=24802380

ralphbean:
This is in place now in Fedora staging -> https://pdc.stg.fedoraproject.org

PR #533, opened 2018-02-05
Sigkey's name shouldn't be null

xiangge:
JIRA: PDC-2562

### DIFF
--- a/pdc/apps/compose/lib.py
+++ b/pdc/apps/compose/lib.py
@@ -197,7 +197,7 @@ def compose__import_rpms(request, release_id, composeinfo, rpm_manifest):
                     imported_rpms += 1
                     path, filename = os.path.split(rpm_data['path'])
                     rpm_id = get_or_insert_rpm(rpms_in_db, cursor, rpm_nevra, srpm_nevra, filename)
-                    sigkey_id = common_models.SigKey.get_cached_id(rpm_data["sigkey"], create=True)
+                    sigkey_id = common_models.SigKey.get_cached_id(rpm_data["sigkey"])
                     path_id = models.Path.get_cached_id(path, create=True)
                     content_category = rpm_data["category"]
                     content_category_id = repository_models.ContentCategory.get_cached_id(content_category)
--- a/pdc/apps/common/migrations/0004_auto_20180205_0922.py
+++ b/pdc/apps/common/migrations/0004_auto_20180205_0922.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('common', '0003_auto_20180131_1318'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='sigkey',
+            name='name',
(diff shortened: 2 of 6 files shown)


lubomir:
I think requiring the name is fine. How will the migration work if there already are keys without name though? Can it create some default name? Maybe reusing key id?

xiangge:
@lubomir As the keyid and name are paired, and it will hard codes in migration file once we set a default, I think we can't set a default value although it gave recommended messages when I do the "makemigrations". Currently I am thinking to update the data manually from admin as there are just several column here.

lubomir:
I agree that updating the data so that correct names are there is the best thing to do, but still maybe having a migration that would deal with incorrect data might be useful. If it explodes in the middle of the update, if might be problematic to access the web admin.

Can we actually do this change? Making a key required when previously it was optional breaks the stability of the API, no?

pdc/apps/compose/lib.py
```diff
@@ -197,7 +197,7 @@ def compose__import_rpms(request, release_id, composeinfo, rpm_manifest):
                     imported_rpms += 1
                     path, filename = os.path.split(rpm_data['path'])
                     rpm_id = get_or_insert_rpm(rpms_in_db, cursor, rpm_nevra, srpm_nevra, filename)
-                    sigkey_id = common_models.SigKey.get_cached_id(rpm_data["sigkey"], create=True)
+                    sigkey_id = common_models.SigKey.get_cached_id(rpm_data["sigkey"])
```
lubomir: Do you know what happens if you try to import compose with new sigkey and it does not exist yet? Is the error message good enough to explain what went wrong?

xiangge: I didn' t try the compose, but I try the release endpoint, it will raise error like below.
{"sigkey":["Object with key_id=abcd12371111 does not exist."]}
So I think this error is enough to understand, and so the same as "compose" endpoint.
What's your thinking?

lubomir:
I still think it would be a good idea to have a migration that would set all missing names to be equal to key IDs. Otherwise the upgrade might get tricky, and we can't update the key everywhere (like Fedora instance), and in some places we just don't really care (like dev machine).

xiangge:
@lubomir I also though about that before, but after I saw the et's sigkey table, I found the name is unique one and not null. So we can't set default value here, because one key_id corresponds one name.

So we just can do it manually to correct the sigkey table with names before upgrade. And we can do anytime because not dependency in codes.

lubomir:
No matter how migration is handled, merging this PR would break the API stability: we can't start requiring a previously optional field.

xiangge:
@lubomir  yep, this changes maybe not quite save, but as I said there are maybe less than 10 items in stage/prod db. So we can first change these data manually, then do the upgrade. 
And if there is a new data inserted or a new pdc server built, will not cause problem because you can just insert one sigkey with requiring name, that's my idea. Or do you have other ideas?

@hluk what's your thinking?

BTW: as I mentioned before the name is unique, so not set a default value here.

xiangge:
@lubomir and @hluk after discuss with you guys, the stability problem can be ignored.
And also without this change, the integration between et and pdc can still work.
We can close this pr.

hluk:
Thanks @xiangge. Hopefully this won't be problematic for the integration in future.

PR #536, opened 2018-02-11
Set default sigkey if user doesn't insert one

xiangge:
Set default sigkey if user doesn't insert one


pdc/settings_local.py.dist
```diff
@@ -84,6 +84,9 @@ EMAIL_SUBJECT_PREFIX = '[PDC]'
 # Omit sending e-mails about big data change when author is in this whitelist.
 #BIG_CHANGE_AUTHORS_WHITELIST = ('some_user',)
 
+# Set the key_id of sigkey "redhatrelease2" as default
```
lubomir: It's not a `key_id`, but name of the key. Also maybe a dummy value would be better as an example (e.g. `Gold Key` or something)?

pdc/settings_test.py
```diff
@@ -48,3 +48,6 @@
 }
 
 COMPONENT_BRANCH_NAME_BLACKLIST_REGEX = r'^epel\d+$'
+
+# Set the key_id of sigkey "redhatrelease2" as default
```
lubomir: The comment should say `name` of key, not `key_id`.

pdc/apps/release/serializers.py
```diff
@@ -134,17 +135,31 @@ class Meta:
...
+        try:
+            if not validated_data["sigkey"]:
+                validated_data["sigkey"] = SigKey.objects.get(name=settings.RELEASE_DEFAULT_SIGKEY)
+        except AttributeError:
+            # If there is no Settings.RELEASE_DEFAULT_SIGKEY, raise it in server log
+            raise AttributeError('Settings.RELEASE_DEFAULT_SIGKEY does not exist, please set one".')
+        except Exception:
+            # If there is no the default SigKey object, just keep it as null.
+            pass
+        return validated_data
```
lubomir: If I understand correctly:
 * if the config option is missing, an error will be reported to users
 * if the config specifies a key that does not exist, nothing will happen

To me it this seems unintuitive: I would expect to be told when I make a typo in the configuration and the key does not exist. Also when the config is not given, it might work similarly to how it did before.

pdc/settings_local.py.dist
xiangge: Here will get the Sigkey object by the default key name, this is also set in et side.
And the name is also unique and not null. And we should keep the sigkey table with the default key_id/name before create a release.

pdc/settings_test.py
xiangge: Yes, will change it.

pdc/apps/release/serializers.py
xiangge: Yep, correct, should change like this.

xiangge:
Commit new changes, please help to review. And this pr is based on #533 

@lubomir User can still insert the data without sigkey, so nothing needs change in ct's input data or scripts.
The default sigkey will set as it is config in the settings file, and also the default key must include in the db first, otherwise it will raise error.

lubomir:
No matter how migration is handled, merging this PR would break the API stability: we can't start requiring a previously optional field.

pdc/settings_test.py
```diff
@@ -48,3 +48,6 @@
 }
 
 COMPONENT_BRANCH_NAME_BLACKLIST_REGEX = r'^epel\d+$'
+
+# Set the name of sigkey as default
```
hluk: Remove the comment.

pdc/apps/release/tests.py
```diff
@@ -1196,7 +1207,7 @@ def test_update_can_explicitly_erase_optional_field(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertIsNone(response.data.get('dist_git'))
         self.assertIsNone(models.Release.objects.get(release_id='release-1.0').dist_git_branch)
-        self.assertNumChanges([1])
+        self.assertNumChanges([2])
```
hluk: What's the additional change about?

pdc/settings_local.py.dist
```diff
@@ -84,6 +84,9 @@ EMAIL_SUBJECT_PREFIX = '[PDC]'
 # Omit sending e-mails about big data change when author is in this whitelist.
 #BIG_CHANGE_AUTHORS_WHITELIST = ('some_user',)
 
+# Set the name of sigkey "redhatrelease2" as default
+RELEASE_DEFAULT_SIGKEY = "redhatrelease2"
```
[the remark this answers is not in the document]
xiangge: If you mean "redhatrelease2", this is the name which is set in ET, can't change.
And as another pr is canceled, here it should be the id not a name. Will change it.
In my understanding, this sigkey is used for ET, so we should keep it same with ET.

pdc/settings_test.py
xiangge: Have already removed that.

pdc/settings_local.py.dist
```diff
@@ -84,6 +84,9 @@ EMAIL_SUBJECT_PREFIX = '[PDC]'
 # Omit sending e-mails about big data change when author is in this whitelist.
 #BIG_CHANGE_AUTHORS_WHITELIST = ('some_user',)
 
+# Set the name of sigkey "redhatrelease2" as default
```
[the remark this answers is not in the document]
xiangge: Added it in the comment

pdc/apps/release/tests.py
xiangge: I was also confused here. After I print the changeset, I really saw two changes.
Here it is:
[{'old_value': u'{"release_id": "release-1.0", "product_version": null, "integrated_with": null, "sigkey": null, "active": true, "allow_buildroot_push": false, "short": "release", "name": "Test Release", "allowed_debuginfo_services": [], "version": "1.0", "allowed_push_targets": [], "base_product": null, "release_type": "ga"}', 'target_id': 1, 'new_value': u'{"release_id": "release-1.0", "product_version": null, "integrated_with": null, "sigkey": "ABCDEF", "active": true, "allow_buildroot_push": false, "short": "release", "name": "Test Release", "allowed_debuginfo_services": [], "version": "1.0", "allowed_push_targets": [], "base_product": null, "release_type": "ga"}', 'target_class': u'release', u'changeset_id': 1, u'id': 1}, 
{'old_value': u'{"dist_git_branch": "release_branch", "release_id": "release-1.0"}', 'target_id': 1, 'new_value': u'null', 'target_class': u'releasedistgitmapping', u'changeset_id': 1, u'id': 2}]

I think the original one is just dist_git_branch change, but after set the default sigkey, here is one release change.

pdc/settings_local.py.dist
```diff
@@ -24,6 +24,8 @@
 # checks, uncomment this line, set DISABLE_RESOURCE_PERMISSION_CHECK with True.
 #     DISABLE_RESOURCE_PERMISSION_CHECK = True
 #
+# Example 3: set the id of sigkey which has name "redhatrelease2" as default
```
hluk: From this description I still wouldn't know what the option does.

pdc/apps/release/serializers.py
```diff
@@ -134,18 +135,32 @@ class Meta:
                   'compose_set', 'integrated_with', 'sigkey', 'allow_buildroot_push',
                   'allowed_debuginfo_services', 'allowed_push_targets')
 
+    def set_default_sigkey(self, validated_data):
+        try:
+            if "sigkey" in validated_data and not validated_data["sigkey"]:
+                validated_data["sigkey"] = SigKey.objects.get(key_id=settings.RELEASE_DEFAULT_SIGKEY)
+        except AttributeError:
+            # If there is no Settings.RELEASE_DEFAULT_SIGKEY, raise it in server log
+            raise AttributeError('Settings.RELEASE_DEFAULT_SIGKEY does not exist, please set one with the Sigkey key_id.')
```
hluk: In a comment you say:

> User can still insert the data without sigkey, so nothing needs change in ct's input data or scripts.

but it doesn't seem to be possible here.

Can you also add tests for that?

Note: You can override settings in Django [like this](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#overriding-settings).

xiangge: This try/catch just checks if the default key includes in db. 
And it means if people not insert one sigkey, will set a default one, so the users' scripts don't need change, they can still add the release without a sigkey input.

pdc/settings_local.py.dist
xiangge: Add more comment

pdc/apps/release/serializers.py
hluk: But it requires `RELEASE_DEFAULT_SIGKEY` option to be set and refer to an existing sigkey. Right? This will require manual update of PDC server's configuration and data after upgrading to a new version.

Also, I don't understand where the default values ("redhatrelease2", "fd431d51") come from. Any specific data shouldn't be pushed to this repository.

xiangge: Yes, the default sigkey should be included into the db first.
And the default sigkey shouldn't change once it set even doing the upgrade.  This  ("redhatrelease2", "fd431d51") is set as default in ET's settings. And also this is already includes in the ET's database.  So we should keep the data same as ET's.

hluk: What about Fedora? Those values should be in ansible repo; `pdc/settings_local.py.dist` is just template for the configuration file, ideally, it should also serve as documentation/reference for the options.

xiangge: aha, good question, I forgot one thing, if we close the pr about the sigkey name's changes, we can let SigKey.get_cached_id do this.

hluk:
There are still issues in this PR:
1. Specific key IDs should not be pushed in the repo. This should go to respective ansible/infrastructure repos.
2. The new option `RELEASE_DEFAULT_SIGKEY` should be optional -- don't force people to use a sigkey when creating releases -- so it doesn't break existing API behavior.

xiangge:
@hluk 
Good idea, here in et the pdc's release is same as legacy product-version,  so the release must have a sigkey (a product-version with a sigkey in et). We can set it in ansible playbooks, and RELEASE_DEFAULT_SIGKEY can be optional and not influence the fedora side then.

pdc/apps/release/serializers.py
```diff
@@ -134,18 +135,29 @@ class Meta:
                   'compose_set', 'integrated_with', 'sigkey', 'allow_buildroot_push',
                   'allowed_debuginfo_services', 'allowed_push_targets')
 
+    def set_default_sigkey(self, validated_data):
+        try:
+            if hasattr(settings, 'RELEASE_DEFAULT_SIGKEY'):
+                if "sigkey" in validated_data and not validated_data["sigkey"]:
+                    validated_data["sigkey"], _ = SigKey.objects.get_or_create(key_id=settings.RELEASE_DEFAULT_SIGKEY)
+        except Exception, e:
+            raise Exception('Error(%s) when get Key_id(%s).' % (e, settings.RELEASE_DEFAULT_SIGKEY))
```
hluk: Is the `except` block needed? What can happen here?

pdc/settings_local.py.dist
```diff
@@ -24,6 +24,10 @@
 # checks, uncomment this line, set DISABLE_RESOURCE_PERMISSION_CHECK with True.
 #     DISABLE_RESOURCE_PERMISSION_CHECK = True
 #
+# Example 3: set the sigkey's key_id fd431d51 as default, this default key_id is
+# used when insert a release into database but without a sigkey, it will be set
+# into the release. This default sigkey's name is "redhatrelease2".
+#     RELEASE_DEFAULT_SIGKEY = "fd431d51"
```
hluk: Remove the key IDs.

I don't like the examples here are used to describe options, it's bit difficult to read. Can you move out and use similar format as for options below (`DEBUG_TOOLBAR`, `BIG_CHANGE_AUTHORS_WHITELIST`)?

It doesn't describe what happens if the key doesn't exist.

xiangge: I am not sure if I can get your point here, I have committed patch for the ansible codes, to set the RELEASE_DEFAULT_SIGKEY there.
And this comment is for other people like fedora guys who want to build a pdc server, they can choose if to set this RELEASE_DEFAULT_SIGKEY.

pdc/apps/release/serializers.py
hluk: You can say that about a lot of code. It actually raises different type of exception here which can change how Django reacts.

So I would rather if you remove it.

pdc/settings_local.py.dist
hluk: I understand, but why mention these particular key IDs?

xiangge: Aha, this is et's default key, as I have set it in ansible codes, not sure if we can remove this to set a fake one, so keep it here for the review :)

pdc/apps/release/serializers.py
xiangge: Make sense to me.

pdc/settings_local.py.dist
hluk: I think it's OK to use the same value as for tests (provided the option is commented-out).

xiangge: Have changed the name to a non-existent one "Sigkey's key_id"

hluk:
@lubomir Can you review?

xiangge:
@lubomir can you help to review it?

pdc/apps/release/tests.py
lubomir: This is correct: dist git mapping is stored in a separate model, and now also the release itself is changing.

pdc/settings_local.py.dist
```diff
@@ -24,6 +24,10 @@
 # checks, uncomment this line, set DISABLE_RESOURCE_PERMISSION_CHECK with True.
 #     DISABLE_RESOURCE_PERMISSION_CHECK = True
 #
+# Example 3: when you insert a release into database without a sigkey data, the sigkey
+# will be null by default, you can also choose if set the key_id as default, this default
+# key_id will be added into the data when insert a releaseinto database without a sigkey.
+#     RELEASE_DEFAULT_SIGKEY = "Sigkey's key_id"
```
lubomir: Minor typo here *releaseinto* → *release into*.

xiangge: Thanks for the review :)

PR #538, opened 2018-02-15
Create resources automatically

lubomir:
Instead of the magic query parameter, we now have a django-admin command that updates the list of resources. This should run after every server update to ensure the list is up-to-date.

The code is basically just moved to a different location. The only change is removing a hardcoded list of actions.

JIRA: PDC-2538


pdc/apps/auth/views.py
```diff
@@ -737,28 +737,6 @@ class ResourcePermissionViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
     filter_class = filters.ResourcePermissionFilter
     API_WITH_NO_PERMISSION_CONTROL = set(['auth/token', 'auth/current-user'])
```
hluk: Is this still used?

pdc/apps/auth/management/commands/updateresources.py
```diff
@@ -0,0 +1,44 @@
...
+from pdc.apps.auth.models import ResourcePermission, ActionPermission, Resource
+from pdc.apps.utils.SortedRouter import router
+from pdc.apps.utils.utils import convert_method_to_action
+
+
+class Command(BaseCommand):
+    help = "Updates list of resources for which permissions can be created"
+    API_WITH_NO_PERMISSION_CONTROL = set(['auth/token', 'auth/current-user'])
+
+    def handle(self, *args, **options):
```
hluk: Is it possible to call this automatically on `pre_migrate` signal rather than creating new command?

pdc/apps/auth/views.py
lubomir: Good point, I missed this one.

pdc/apps/auth/management/commands/updateresources.py
lubomir: Good suggestion. It might be possible. I'll need to make sure the signal is fired even if there are no actual migrations to run.

lubomir: It works, the signal is fired even if there is no change. I used `post_migrate` to make sure that the database schema matches the models. Would `pre_` be better?

hluk: I guess `post_` is OK, not sure if any migrations would want to access it.

hluk:
@lubomir Did you notice that Travis fails?

lubomir:
The tests are failing because now there are resource permission created during testing, and the test code does not turn permission checking off. So far it worked because with no permissions the check did not do anything.

I'll try to fix it and update the PR.

lubomir:
There was a problem with the tests that we didn't notice before: when a test changed some setting, it was not always correctly restored. Now all tests should be passing.

pdc/apps/auth/apps.py
```diff
@@ -4,9 +4,14 @@
...
 
 
 class AuthConfig(AppConfig):
     name = 'pdc.apps.auth'
     label = 'kerb_auth'
     verbose_name = 'Auth with Kerberos support'
+
+    def ready(self, *args, **kwargs):
+        from . import signals
+        post_migrate.connect(signals.update_resources, sender=self)
```
hluk: Is it be possible to use `@receiver(post_migrate)` decorator on `update_resources()` instead?

pdc/apps/auth/tests.py
```diff
@@ -337,17 +342,17 @@ def test_resource_permission_all_read_permissions_on(self):
             for permission in ActionPermission.objects.all():
                 ResourcePermission.objects.create(resource=resource, permission=permission)
 
-        settings.ALLOW_ALL_USER_READ = False
-        response = self.client.get(reverse('currentuser-list'), format='json')
+        with self.settings(ALLOW_ALL_USER_READ=False):
```
hluk: Nice. This looks a lot safer than changing settings on random places.

pdc/apps/auth/signals.py
```diff
@@ -0,0 +1,40 @@
+#
+# Copyright (c) 2018 Red Hat
+# Licensed under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+#
+
+
+def update_resources(sender, **kwargs):
```
hluk: Could be skipped for tests by adding a new option to `settings_test.py` and checking the value here (deleting these objects in tests won't be needed).

This function can be also tested, simply by calling this function and checking that a known resource is available.

pdc/apps/auth/apps.py
lubomir: It's doable, but requires further code to make sure it runs only once. By default `@receiver` would connect it to all senders, so the code would run for each application that has migrations. Given that we should still import the `signals` module in `ready()` method, I does not seem like a big win.

hluk: OK, understood.

lubomir:
The failed tests are caused by mistakenly pushing the branch to the main repo instead of my fork. I deleted it, but Travis already started running tests on it.

PR #543, opened 2018-02-28
Dummy messenger should log the messages

lubomir:
Instead of ignoring them all, it should log them. This can be easily disabled in the LOGGING settings by adding following snippet:

    'pdc.apps.utils.messaging': {
        'level': 'WARNING',
    }

JIRA: PDC-2583

### DIFF
--- a/pdc/apps/utils/messaging.py
+++ b/pdc/apps/utils/messaging.py
@@ -13,11 +13,9 @@ logger = logging.getLogger(__name__)
 
 
 class DummyMessenger(object):
-    def __init__(self):
-        pass
 
     def send_message(self, topic, msg):
-        pass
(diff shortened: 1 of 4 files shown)


pdc/apps/utils/messaging.py
```diff
@@ -14,11 +14,9 @@
 
 
 class DummyMessenger(object):
-    def __init__(self):
-        pass
 
     def send_message(self, topic, msg):
-        pass
+        logger.info('Sending to %s:\n%s' % (topic, json.dumps(msg, sort_keys=True, indent=2)))
```
hluk: Isn't `msg` already string?

lubomir: Right, it is. That will have to change probably.

lubomir:
I added another commit that encodes the message to JSON only in the messenger. This saves us the parsing in Fedmsg, and will be needed for UMB to construct the headers based on message data.

PR #544, opened 2018-03-01
POC on sending messages for all changes

lubomir:
The first commit adds a mixin that detects the changes and publishes messages, the second one enables it for a couple resources for testing.

There are a couple of things that need addressing:

- [ ] generating the urls could probably be done better
- [ ] testing to find out if this works for viewsets that don't support the full range of actions
- [ ] documentation


pdc/apps/common/viewsets.py
```diff
@@ -63,6 +65,52 @@ def perform_create(self, serializer):
...
+        serializer_class = self.get_serializer_class()
+        self.initial_status = serializer_class(instance=obj,
+                                               context={'request': self.request}).data
+        return obj
+
+    def _get_url(self, basename):
+        if not hasattr(self, 'object') or not self.object:
+            return None
+        # TODO how reliable is this?
+        return reverse('%s-detail' % basename, args=[getattr(self.object, self.lookup_field)])
```
hluk: Probably unreliable, there can be more lookup fields.

Can you use [self.request.build_absolute_uri()](https://docs.djangoproject.com/en/1.8/ref/request-response/#django.http.HttpRequest.build_absolute_uri)?

lubomir: That works for update, but not for create we would get the url for the whole collection, not for the new object.

hluk: This is basically what I try to use for auto-generating API documentation:

```python
def lookup_field_names(view):
    if hasattr(view, 'lookup_fields'):
        return [field for field, _ in view.lookup_fields]

    if hasattr(view, 'lookup_field'):
        return [view.lookup_field]

    return []
```

pdc/apps/common/viewsets.py
```diff
@@ -63,6 +65,64 @@ def perform_create(self, serializer):
...
+        self.initial_status = serializer_class(instance=obj,
+                                               context={'request': self.request}).data
+        return obj
+
+    def _get_url(self, basename):
+        if not hasattr(self, 'object') or not self.object:
+            return None
+        if hasattr(self, 'lookup_fields'):
+            lookup_field_values = [get_related_attr(self.object, f) for f, _ in self.lookup_fields]
+            lookup_field = '/'.join(lookup_field_values)
```
hluk: Won't work for `releases/{release_id}/rpm-mapping/{package}/`.

This is bit crazy, but could help:
```python
from django.core.urlresolvers import get_resolver
resolver = get_resolver(None)
resolver.reverse_dict.getlist('releaserpmmapping-detail')[1][0][0] # This is bit random :)
# (u'rest_api/v1/releases/%(release_id)s/rpm-mapping/%(package)s/', [u'release_id', u'package'])
```

No idea if there is simple way to get URL. :(

pdc/apps/common/viewsets.py
lubomir: This is pretty much what I ended up doing in the last commit. Having simple numeric IDs sure looks much more pleasant…

pdc/apps/common/viewsets.py
```diff
@@ -63,6 +65,64 @@ def perform_create(self, serializer):
...
+            lookup_field = getattr(self.object, self.lookup_field)
+        rel_url = reverse('%s-detail' % basename, args=[lookup_field])
+        return self.request.build_absolute_uri(rel_url)
+
+    def _send_message(self, action, msg):
+        for resource, cls, basename in router.registry:
+            if isinstance(self, cls):
+                break
+        else:
+            raise RuntimeError('What do we do now? Is it even possible?')
```
hluk: I would suspect that if this happens there is no route to the view and therefore it must have been changed internally.

pdc/apps/common/viewsets.py
```diff
@@ -63,6 +65,64 @@ def perform_create(self, serializer):
...
+            lookup_field_values = [get_related_attr(self.object, f) for f, _ in self.lookup_fields]
+            lookup_field = '/'.join(lookup_field_values)
+        else:
+            lookup_field = getattr(self.object, self.lookup_field)
+        rel_url = reverse('%s-detail' % basename, args=[lookup_field])
+        return self.request.build_absolute_uri(rel_url)
+
+    def _send_message(self, action, msg):
+        for resource, cls, basename in router.registry:
+            if isinstance(self, cls):
```
hluk: This could break when sub-classing another view. Maybe `type(self) is cls`?

pdc/apps/common/viewsets.py
lubomir: Would an internal change be coming via the viewset?

pdc/apps/common/viewsets.py
lubomir: I'm not seeing how this could happen? Do you mean there would be a subclass of a viewset available under different route?

pdc/apps/common/viewsets.py
hluk: I think this probably cannot happen, so keeping an exception here is fine.

pdc/apps/common/viewsets.py
hluk: Yes, I guess it's possible to subclass an existing view (that already has route) to get slightly different functionality.

lubomir:
Let's list the problematic resources:

* `compose-images`, `compose-rpms` – these import big amounts of data; and are already handled separately
* `composes` – deleting is really just updating a field, and will need to be handled separately
* `composes/{}/rpm-mapping` – this is tricky one, it has different semantics for `PUT` and `PATCH`, and feels a bit like RPC; maybe it should be handled manually (it should probably send messages about created or deleted RPM overrides). Editing RPM overrides via `overrides/rpm` works fine with the mixin. Editing via web interface would need to be handled manually.
* `compose-tree-locations` – these can be created directly (which the mixin can handle), or via compose full import (which would have to be handled manually)
* `global-components/{}/labels` – can not use mixin because it has custom logic for creating objects; requires manual handling
* all `rpc/*` things will need manual handling
* `RPMRelatedField` can create new RPMs, it should also send messages about them.

These are fine:

* `*-rtt-tests`, `osbs` – they are created automatically when the corresponding entity is created, so there will never be messages about creation and deletion, but updates work with the mixin

lubomir:
I merged latest `master` into this PR (but will rebase and clean up history before merge). There are now comments explaining more stuff, and a bit of clean up in the code.

pdc/apps/common/viewsets.py
```diff
@@ -63,6 +65,86 @@ def perform_create(self, serializer):
...
+        else:
+            # This should never happen: when the viewset is called, the request
+            # needs the mapping in router to actually make it to the current
+            # viewset.
+            raise RuntimeError('Viewset %r not exposed in router.' % self)
+        topic = '.%s.%s' % (resource, action)
+        msg.update({
+            'url': self._get_url(obj) if obj else None,
+            'author': self.request.user.username,
+            'comment': self.request.META.get("HTTP_PDC_CHANGE_COMMENT", None),
```
hluk: Do we also want to send date/time of the change?

pdc/apps/common/tests.py
```diff
@@ -673,3 +676,119 @@ def test_ordering_bad_key(self):
...
+            request._stream = BytesIO(data)
+        return request
+
+    @mock.patch('pdc.apps.common.viewsets.reverse')
+    def test_create_notification(self, mock_reverse):
+        request = self._mk_request('POST', {'nickname': 'Rover', 'color': 'brown'},
+                                   comment='Sit!')
+        mock_reverse.return_value = '/dogs/2/'
+
+        with mock.patch('pdc.apps.common.viewsets.router') as mock_router:
```
hluk: Can you also use decorator mock.patch for this?

pdc/apps/common/tests.py
```diff
@@ -673,3 +676,119 @@ def test_ordering_bad_key(self):
...
+            serializer_class = DummySerializer
+
+        mapping = {
+            'post': 'create',
+            'put': 'update',
+            'delete': 'destroy',
+        }
+        self.view = DummyView.as_view(mapping, basename='dummy')
+
+    def _mk_request(self, method, data=None, comment=None):
```
hluk: mk -> make

pdc/apps/common/viewsets.py
```diff
@@ -63,6 +65,86 @@ def perform_create(self, serializer):
...
+
+        self._send_message('removed', {'old_value': self.initial_status})
+
+    def perform_update(self, serializer):
+        super(NotificationMixin, self).perform_update(serializer)
+
+        if self.initial_status != serializer.data:
+            self._send_message('changed',
+                               {'old_value': self.initial_status,
+                                'new_value': serializer.data},
```
hluk: Sending just the changed fields here would be nice (non-recursively, only the top-level fields).

pdc/apps/common/viewsets.py
lubomir: Yes, and possibly also the changeset id.

pdc/apps/common/tests.py
lubomir: Good suggestion.

pdc/apps/common/viewsets.py
lubomir: I think having all the data is better: it means consumers don't have to do a query if they need also some of the unchanged fields.

hluk: I meant that `new_value` doesn't have to contain all fields if the values are same as in `old_value`.

lubomir: Oh, that makes sense. I created a ticket PDC-2629 for this. I think it could be done in phase 2.

hluk: OK, good, I think it should be easy fix.

lubomir:
I rebased the commits so to have nice history and added the mixin to all viewsets, not just the example ones.

PR #545, opened 2018-03-01
Fix search forms for base product and product version

lubomir:
They should search in the full ID. Since it contains short and version, we can remove those two fields.

PDC-2478

### DIFF
--- a/pdc/apps/release/forms.py
+++ b/pdc/apps/release/forms.py
@@ -28,8 +28,6 @@ class ReleaseSearchForm(forms.Form):
             query |= Q(name__icontains=search)
             query |= Q(release_id__icontains=search)
 
-        # TODO: disabled
-
         return query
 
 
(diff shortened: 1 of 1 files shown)


pdc/apps/release/forms.py
```diff
@@ -43,9 +43,9 @@ def get_query(self, request):
         query = Q()
 
         if search:
-            query |= Q(short__icontains=search)
-            query |= Q(version__icontains=search)
             query |= Q(name__icontains=search)
+            query |= Q(base_product_id__icontains=search)
+            query |= Q(release_type__short__icontains=search)
 
         # TODO: disabled
```
hluk: Oops, some leftover comment. :)  (Blame is of course the "initial import" commit.)

lubomir: This was probably meant to allow searching by whether it's active. That is possible by filtering the table, so I'll add another commit to remove these comments.

PR #551, opened 2018-03-09
Send messages from RPMRelated field

lubomir:
When it creates a new object in database, it should publish a message about it.

Both use cases are updated to test that the message was actually sent.

JIRA: PDC-2603

### DIFF
--- a/pdc/apps/module/tests.py
+++ b/pdc/apps/module/tests.py
@@ -10,6 +10,7 @@ from rest_framework import status
 
 from pdc.apps.common.test_utils import TestCaseWithChangeSetMixin
 from pdc.apps.module.models import Module
+from pdc.apps.utils import messenger
 
 
 class ModuleAPITestCase(TestCaseWithChangeSetMixin, APITestCase):
@@ -100,7 +101,13 @@ class ModuleAPITestCase(TestCaseWithChangeSetMixin, APITestCase):
             'srpm_name': 'foobar',
             'srpm_commit_branch': 'master'
         }]
-        response = self.client.post(url, self.data, format='json')
+        with messenger.listen() as messages:
+            response = self.client.post(url, self.data, format='json')
+
+            # Check two messages were sent.
+            self.assertEqual([m[0] for m in messages],
+                             ['.rpms.added', '.modules.added'])
+
         # Check the response
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         self.expected['rpms'] = ['foobar-0:1.0.0-1.src.rpm']
@@ -109,6 +116,7 @@ class ModuleAPITestCase(TestCaseWithChangeSetMixin, APITestCase):
         url_two = reverse('modules-detail', args=[uid_one])
         response_two = self.client.get(url_two, format='json')
         self.assertEqual(response_two.data, self.expected)
+
         # Create another module
         self.data['version'] = '56789012'
         self.data['rpms'] = [{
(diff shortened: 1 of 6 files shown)


pdc/apps/module/tests.py
```diff
@@ -100,7 +101,9 @@ def test_create_and_get_module_with_rpms(self):
             'srpm_name': 'foobar',
             'srpm_commit_branch': 'master'
         }]
-        response = self.client.post(url, self.data, format='json')
+        with messenger.listen() as messages:
```
hluk: Looks kinda strange when `messages` are not accessed withing this block.

pdc/apps/package/serializers.py
```diff
@@ -139,6 +140,16 @@ def to_internal_value(self, data):
                                               rpm.id,
                                               'null',
                                               json.dumps(rpm.export()))
+
```
hluk: The function is pretty large with big cyclomatic complexity. Can you do something about it?

pdc/apps/package/serializers.py
```diff
@@ -139,6 +140,16 @@ def to_internal_value(self, data):
...
+                    if request and hasattr(request, '_messagings'):
+                        msg = {
+                            'url': reverse('rpms-detail', args=[rpm.pk], request=request),
+                            'author': request.user.username,
+                            'comment': request.META.get("HTTP_PDC_CHANGE_COMMENT", None),
```
hluk: Aren't these fields always overriden by `NotificationMixin`?

lubomir: No, the mixin is adding another message. Once the message is in the queue, there is nothing that would change it's content. But it might be a good idea do some post processing somewhere.

hluk: Oh, OK, got it. Would it be possible to move the common code for adding these fields to a helper function?

pdc/apps/package/serializers.py
lubomir: Do you have a tool to measure the complexity? (In this case it's obvious, but having something automatic would be nice.)

hluk: Not for Python - using only my eyes - though there must some other tool :).

lubomir:
All comments should be addressed.

I'm wondering what to do about the `ArchiveRelatedField`. It's also creating new objects, but there is no general API for the archives... I guess we can probably ignore this (there should be no users for the build-images anyway).

pdc/apps/package/serializers.py
```diff
@@ -107,48 +108,74 @@ class Meta:
...
+        try:
+            required_data[field] = data[field]
+        except KeyError:
+            errors[field] = 'This field is required.'
+    if errors:
+        raise serializers.ValidationError(errors)
+    return required_data
+
+
+def _log_new_object(request, obj, route, topic, data):
```
hluk: Would be better to split this into two functions, one creating changeset and other the message.

pdc/apps/package/serializers.py
```diff
@@ -107,48 +108,74 @@ class Meta:
...
+                                       context={'request': request})
+            if not serializer.is_valid():
+                raise serializers.ValidationError(serializer.errors)
+
+            rpm = serializer.save()
+            _log_new_object(request, rpm, 'rpms-detail', 'rpms', serializer.data)
+            return rpm
+        except Exception as err:
+            raise serializers.ValidationError("Can not get or create RPM with your input(%s): %s." % (data, err))
+
```
hluk: Nice. Much easier to read.

PR #552, opened 2018-03-12
Add producer based on rhmsg

lubomir:
There is special code to extract data from the message body to include as headers.

JIRA: PDC-2584

Todo:

* [x] error handling
* [x] tests


lubomir:
This is a very basic version that has a significant problem: it blocks the request until the message is sent, which according to my testing takes at least two seconds (and with default rhmsg timeout it could be over minute). I don't think this is acceptable.

IMO we need to send the messages in the background. I see two options:
 * use a thread that would run in the background
 * bring in the big guns and use celery

pdc/apps/utils/messaging.py
```diff
@@ -124,3 +124,38 @@ def send_message(self, topic, msg):
...
+    def send_message(self, topic, msg):
+        from rhmsg.activemq.producer import AMQProducer
+
+        headers = self._make_headers(msg)
+        topic = '%s%s' % (self.prefix, topic)
+        logger.info('Sending to %s', topic)
+
+        with AMQProducer(**self.config) as producer:
+            producer.through_topic(topic)
+            producer.send_msg(headers, json.dumps(msg))
```
hluk: Looking at the `rhmsg` code, this raises an exception if sending fails. Is it OK not to catch it?

lubomir:
All comments should be addressed now. I also slightly modified the code to make sure we can send multiple messages at once (if they have the same topic).

lubomir:
The last commit adds a very simple solution for the blocking problem: it spawns a background thread and sends the messages to it via a queue. It works quite well in my testing, but there are a few possible issues:

* we don't limit the length of the queue, so it could grow indefinitely (and we don't have any monitoring...)
* if the server crashes while there are unsent messages, they will be lost

I don't think there will be enough messages for the queue to actually grow much, so these are probably not very serious issues.

I'll add tests if we decide this is the way we want to go. Alternative solutions could be [django background tasks](http://django-background-tasks.readthedocs.io/en/latest/) (which puts tasks into database and invokes them from cron) or celery (which would require a separate worker process and redis or rabbitmq).

pdc/apps/utils/messaging.py
```diff
@@ -179,21 +185,23 @@ def send_messages(self, msgs):
         This can save a lot of time for bulk changes since there will be only
         one connection per group.
         """
-        for group in itertools.groupby(msgs, lambda x: x[0]):
-            self._send_group(*group)
+        for topic, msgs in itertools.groupby(msgs, lambda x: x[0]):
+            self.queue.put((topic, list(msgs)))
```
hluk: Can you call Queue.join() when httpd is being stopped (e.g. before upgrade)?

lubomir: I'm not sure, there does not seem to be any signal or hook that could be used for that. Is it useful though? What would it achieve?

I'm wondering whether using `multiprocessing` would be better than the thread.

hluk: I'm not sure what happens with the thread when httpd is stopped.

lubomir: I would expect the main thread that runs the application to be stopped, and since this thread is marked as `daemon`, it should be killed (resulting in loss of any unsent messages still in queue).

hluk: Well, there could be also some network issues so the reliability of actually delivering the messages could be resolved later if needed (using RabbitMQ would be nice but will probably take a while to implement with fixing all ansible scripts).

hluk:
Do you plan to add tests to this PR?

lubomir:
Yes for tests, I'm currently working on moving the code to a separate django application.

pdc/apps/messaging/apps.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def ready(self):
+        backend = settings.MESSAGE_BUS.get('BACKEND')
+        if backend:
+            cls = import_string(backend)
+            self.messenger = cls()
+        else:
+            # Drop this block in next version and instead directly fall on
+            # DummyMessenger.
+            from backends import capture, dummy, fedmsg, rhmsg
```
hluk: Oh, looks like `django-admin makemigrations --check` calls `ready()` for apps.

It's perhaps better to make `MESSENGERS` map from the known strings to dotted path and use same code as above.

PR #559, opened 2018-03-23
Fix form validation in web UI

lubomir:
The `is_valid` method only returns a bool indicating success or failure, there are no exceptions. This can actually lead to a 500 error.

The error can be replicated with

    curl -X HEAD https://pdc/compose/<pk>/images/<variant>/<arch>

with valid combination of compose pk, variant uid and arch.


lubomir:
I found this by running following command to see what errors are printed when automatically generating docs for API.

    wget -r --spider --header 'Accept: text/html' http://127.0.0.1:8000/rest_api/v1/ --wait=0.5

